Issue #66, opened 2014-10-18
Add a Mode aggregate function.


AliTajeldin:

ninjapapa:
For this requirement, does the mode on a String field or a numeric field or general?

AliTajeldin:
Looks good.

PR #345, opened 2016-08-19
Feature/mode update

abhikesh:
Simple version of Mode (#66). Can also be called Most Frequent Value. Extends Histogram to find the most frequent values in the histogram - returns multiple modes. There could be extreme cases (think uniform discrete distribution in the limiting case) where the number of modes are too many; in that case we will want to limit the number of modes returned.


Kai-Chen:
Thanks for the PR!  I've merged the feature and skipped whitespace changes, as it distracts from what is essential.  Needs a bit more tests, but I can add that.

Many thanks again for the contribution!

abhikesh:
Thank you. In that case, would you like to close #66 as well? Are there any other features that you want from the Mode function?

Kai-Chen:
I will check with @AliTajeldin to see if he still remembers what that ticket is about; it's been almost 2 years since it was opened.

PR #583, opened 2017-04-03
warnings/typos on 2.1

hubertp-lshift:

PR #587, opened 2017-04-04
Fix Python names in docs 

hubertp-lshift:
(followup on #583)


hubertp-lshift:
That's all, I'm not gonna push more things into this PR (since it is only a follow up on #583).

Issue #517, opened 2017-03-10
Add code formatting to sbt builds


hubertp-lshift:
https://github.com/olafurpg/scalafmt

hubertp-lshift:
Fixed in #592 and #593. Need to ensure that it stays that way in #594.

PR #593, opened 2017-04-05
Applied scalafmt to 2.1

hubertp-lshift:
Same as https://github.com/TresAmigosSD/SMV/pull/592 except that applied to 2.1.
Since the formatting is consistent, it should ensure that the merges from master are painless.


hubertp-lshift:
This refers to issue https://github.com/TresAmigosSD/SMV/issues/517 but on 2.1 branch.

PR #621, opened 2017-04-12
Update smv-install to include latest release version in usage instructions

jacobdr:
Provide slightly more helpful usage message that includes the version number of the most recent SMV release

### DIFF
--- a/tools/smv-install
+++ b/tools/smv-install
@@ -6,7 +6,7 @@ PROG_NAME=$(basename "$0")
 function usage()
 {
   echo "USAGE: ${PROG_NAME} [-spark] smv_version_to_release(n.n.n.n)"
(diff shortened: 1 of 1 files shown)


tools/smv-install
```diff
@@ -6,6 +6,8 @@ PROG_NAME=$(basename "$0")
 function usage()
 {
   echo "USAGE: ${PROG_NAME} [-spark] smv_version_to_release(n.n.n.n)"
+  LATEST_VERSION=$(curl -s https://api.github.com/repos/tresamigossd/smv/releases | grep browser_download_url | head -n 1 | cut -d '"' -f 4 | cut -d '/' -f 8 | cut -c 2-)
```
AliTajeldin: I think it might be easier and cleaner to just get the contents of the `.smv_version` file from master:
`curl https://raw.githubusercontent.com/TresAmigosSD/SMV/master/.smv_version`

jacobdr: Completely agreed. Changed in bb17f58

PR #622, opened 2017-04-12
 #618 ported smvStrCat to Python

ninjapapa:
* Scala side, moved smvStrCat to smvfuncs from package, while still kept the version in package (labeled deprecated)
* Implemented smvStrCat on Python side with docstr
* Created test on Python side

### DIFF
--- a/src/main/python/smv/functions.py
+++ b/src/main/python/smv/functions.py
@@ -14,6 +14,7 @@
 from smv import SmvApp
 from pyspark.sql.column import Column
 from pyspark.sql.functions import udf
+from utils import smv_copy_array
 
 def nGram2(c1, c2):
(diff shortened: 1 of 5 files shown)


src/main/python/smv/functions.py
```diff
@@ -128,3 +129,18 @@ def smvCollectSet(col, datatype):
             datatype (DataType): datatype of the input column
     """
     return Column(SmvApp.getInstance()._jvm.org.tresamigos.smv.python.SmvPythonHelper.smvCollectSet(col._jc, datatype.json()))
+
+def smvStrCat(head, *others):
+    """Concatenate multiple columns to a single string. Similar to `concat` and `concat_ws` functions in Spark but behaves differently
```
laneb: Docstrings should describe the argument and return types.

Issue #619, opened 2017-04-11
port smvHashKey to Python


ninjapapa:

ninjapapa:
c973383223cbb3f9b0911c8c8444cd7f44dc3a25

done

PR #624, opened 2017-04-12
#619 smv hash key

ninjapapa:
* Changed smvHashKey's behavior when specified prefix
* Ported smvHashKey to Python

### DIFF
--- a/src/main/python/smv/functions.py
+++ b/src/main/python/smv/functions.py
@@ -132,7 +132,7 @@ def smvCollectSet(col, datatype):
 
 def smvStrCat(head, *others):
     """Concatenate multiple columns to a single string. Similar to `concat` and `concat_ws` functions in Spark but behaves differently
-       when some columns are nulls. 
+       when some columns are nulls.
     """
     if (isinstance(head, basestring)):
         sep = head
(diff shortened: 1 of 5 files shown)


src/main/python/smv/functions.py
```diff
@@ -144,3 +157,46 @@ def smvStrCat(head, *others):
...
+    if (isinstance(head, basestring)):
+        pre = head
+        cols = list(others)
+    elif (isinstance(head, Column)):
+        pre = ""
+        cols = [head] + list(others)
+    else:
+        raise RuntimeError("first parameter must be either a String or a Column")
+    app = SmvApp.getInstance()
+    return Column(app._jvm.org.tresamigos.smv.python.SmvPythonHelper.smvHashKey(pre, smv_copy_array(app.sc, *cols)))
```
jacobdr: If it is trivial switch between md5, sha1, and sha2 on the Spark side, and we are spending time defending md5, am curious to know why we don't just use one of the other hashing algorithms. (md5 use [here](https://github.com/TresAmigosSD/SMV/pull/624/files#diff-57c0956d2193fa9828996ae0dd2c593aR158))

If it is because it is a breaking change for existing users, maybe we should mention that in the developer docstring -- and would be the kind of change we would look to make when we get to 2.0

ninjapapa: There are quite some this type of helpers which were necessary before Spark offering similar functions. May need a review down the road, but not right now. Because of existing projects and existing customers, we'd like to encourage them to move with us to 2.1, we can't simply drop those helpers. Also the intent of SmvHashKey is to provide a way to create an unique key from a group of columns. The fact of using md5 is the implementation of that need. Not all users need to know the details of how an unique key should be implemented. Through this method we made the decision for them. On the other hand power user should feel free to us whatever algorithms as long as Spark support.

Issue #605, opened 2017-04-07
Merge master to 2.1


ninjapapa:
Since scalafmt has been applied to both master and 2.1 branch separately, it is pretty hard to merge master to 2.1. Need to pay the effort and fix this before they move too far.

Kai-Chen:
Apparently it's not just the use of scalafmt.
There are quite a few other undesirable changes that crept into the 2.1 branch, at least from a merging perspective.

Below is a list of the tasks to perform:
- [ ] Revert test-related changes first so we can run tests.
- [ ] Revert commits that introduced file renaming and repackaging, both of which unnecessarily complicates merge.
- [ ] Double check to make sure that we did not leave out the intended changes.

hubertp-lshift:
Since I committed most of those changes, I can have a look at it and make a PR.

hubertp-lshift:
Also, I don't think reverting is a way to go. You will be stuck with some pretty obvious file structure problems that should be fixed. I think the main question is where those changes should be applied. I assumed that the main dev will gradually shift towards 2.1 so merging will be unnecessary, but maybe I'm wrong?

hubertp-lshift:
FYI, instead of reverting some of the changes, I would probably just cherry-pick them, apply to master, merge again into 2.1, which will be painless. It's a longer, but probably a safer route which I can volunteer to do. I will apply the non-2.1 specific changes to master from now on, until there is a decision to switch dev to it.

AliTajeldin:
@hubertp-lshift correct.  master is our dev branch.  We will merge the 2.1 branch into master and create a 1.5.2.x maintenance branch in next couple of days.

Kai-Chen:
The file renaming and repackaging definitely does not belong in this branch.
In fact, doing that task on a devoted branch by itself is probably a good idea.
It would keep simple isomorphic changes separate from functional changes, the stuff that really is different.

So I would say at least revert these changes on the 2.1 branch, and if you feel really strong about the names, send a separate PR for that only, to the master branch, and then we can discuss there.

Ultimately, though, I think these things and stylistic stuff is a time and energy sink.
It's necessary process to deal with the unfortunate fact that, when we read code, we let what the code **looks like** distract us from understanding what the code **means**.

AliTajeldin:
Ok guys, to put an end to the back-and-forth, here is the plan:
* For future code refactors, they need to be on a branch with a separate PR against master and a github issue.
[...]

PR #626, opened 2017-04-12
Sync 2.1 with latest master

hubertp-lshift:
Issue #605


hubertp-lshift:
Closing, as per discussion on the issue #605. I will revert the changes to package renaming so then it should be easier to merge, modulo scalafmt changes which are consistent among the two branches.

PR #632, opened 2017-04-13
Resolve some of the issues on #605 by reverting

hubertp-lshift:
Will reapply on master (at least the package renaming, codebase cleanup)


hubertp-lshift:
Once this change is approved/merged I have another one that merges master into 2.1

Issue #620, opened 2017-04-12
Pass through smv args to `smv-pytest`


jacobdr:
Right now the following code fails:
`smv-pytest --smv-props smv.config.sample=full`

I get this error message:
`ImportError: No module named --smv-props`

laneb:
Should the title of this issue be about `smv-pytest` instead of `smv-pyrun`?

laneb:
In SMV's Python tests, smv args are supplied for each test in its setup - we don't even create an `SmvApp` until the setup step for the first test. If the desired behavior is to combine smv args supplied at the command line with smv args specified in test setup, one option is to give `SmvBaseTest` access to the command line smv args so that they are combined with the test-specific arguments when `SmvBaseTest` initializes `SmvApp`.

jacobdr:
To provide some more context around the motivation...

In our normal development, our project users tend to have a run-time config set in `smv-user-conf.props` that drops them into a <1%  sampled universe. They work on a module, write some tests, and tend to invoke `smv-pytest` with an argument specifying just the test module / suite that is relevant to their immediate work.

Occasionally we want to run `smv-pytest` and enforce that the run-time config be set to "full data" -- especially when operating on our CI box. The easiest way to force this behavior is a simple bash script that wraps `smv-pytest` and provides the `--smv-props` flag with the value that we want. 

Does that make sense @laneb?

laneb:
Yes, that makes sense. Currently, how are you getting an instance of `SmvApp` (we need to make sure your args make it to `SmvApp`)? Are you instantiating it yourself, or are you using `SmvBaseTest`, which instantiates it for you?

laneb:
This appears to be more or less a direct c/p of the SMV's `SmvBaseTest`, though probably a month or two old (`smvPy` is now `smvApp`). If you want to continue using this `SmvBaseTest`, you will have to supply the args to `smvPy` yourself.

I will make the following changes:

- Individual tests will be run with `-t`
- `runtest.py` will only eat `-t` and the argument immediately following it.
- The remaining options will be available in `TestConfig.commandlineArgs`

You will just have to add `TestConfig.commandlineArgs` to the args that are passed to `smvPy.init`.

laneb:
I will also move our `SmvBaseTest` into `src/main/python` so that it is easier for you to import it instead of rolling your own.

laneb:
done.

PR #650, opened 2017-04-19
I620 smvpytest args

laneb:
Addressing #620. @jacobdr please have a look.


src/main/python/smv/helpers.py
```diff
@@ -341,6 +341,10 @@ def smvJoinMultipleByKey(self, keys, joinType = 'inner'):
         jdf = self._jPythonHelper.smvJoinMultipleByKey(self._jdf, smv_copy_array(self._sc, *keys), joinType)
         return SmvMultiJoin(self._sql_ctx, jdf)
 
+    def topNValsByFreq(self, n, col):
```
AliTajeldin: Missing pydoc description

src/main/scala/org/tresamigos/smv/SmvDFHelper.scala
```diff
@@ -389,7 +388,7 @@ class SmvDFHelper(df: DataFrame) {
    *
    * and the following call:
    * {{{
-   *   df.dedupByKey("id")
+   *   df.debupByKey("id")
```
AliTajeldin: why rename `dedupByKey` to `debupByKey`? What is a `debup`?

laneb: I have no idea how this got changed. @Khorkhe at some point mentioned finding some  `debupByKey` typos in the documentation, but I don't know how I could've made this change.

src/main/python/smv/helpers.py
laneb: Must have committed this to both branches I guess. Strange. I'll document `topNValsByFreq` on the skewjoin branch and merge it there.

PR #654, opened 2017-04-20
I497 site doc

Khorkhe:
run `tools/gen-docs.sh` to auto-generate python and scala docs, commit them to gh-pages and update index.html links


AliTajeldin:
merged!

Issue #607, opened 2017-04-08
datasetHash of SmvModuleLink should always vary with target module


laneb:
Given an `SmvModuleLink` created like
```
val L1 = SmvModuleLink(fqn.of.A)
```
as occurs in `DataSetMgr`, if we also create a second module link 
```
val L2 = SmvModuleLink(fqn.of.B)
``` 
the datasetHash of `L1` should be different from the hash of `L2` even if the data for `fqn.of` is versioned. This is not the case and recently led to a confusing situation in which 2 Python `SmvModuleLinks` linking to different modules had the same `datasetHash`.

laneb:
Done.

PR #659, opened 2017-04-21
SmvModuleLink datasetHash should vary with target module (#607)

laneb:
Addressing #607. Because this changes `SmvModuleLinks'` `datasetHash`, I had to update a test that was asserting that an `SmvModuleLink's` `datasetHash` was equal to a hard-coded value. We should find an alternative to these hardcoded values.


AliTajeldin:
looks good and a really nice testcase. merging now.

PR #669, opened 2017-04-24
I646 python loading

laneb:


src/main/python/smv/datasetrepo.py
```diff
@@ -38,42 +38,34 @@ class Java:
...
     def __init__(self, smvApp):
         self.smvApp = smvApp
+        self._clear_sys_modules()
+
+    def _clear_sys_modules(self):
+        """Clear all client modules from sys.modules
+        """
+        for fqn in sys.modules.keys():
+            for stage_name in self.smvApp.stages:
+                if fqn.startswith(stage_name):
```
laneb: This breaks if we skip deleting the stage object - basically we lose the ability to dynamically reload modules. Haven't dug into why, but I think it has to do with how `pkgutil.walk_packages` works. We should be able to make this work for us by deleting modules beginning with `stage_name + '.'` _and_ the module `stage_name`. Do you foresee any issue with that?

Issue #627, opened 2017-04-12
Need an option to add jars to the classpath


laneb:
If you use Spark's `--jars` it is ignored, probably because it's the second --jars option in the resulting `spark-submit` (first one is for the `APP_JAR`). Listing additional jars with the `APP_JAR` works.

laneb:
@AliTajeldin Thoughts on just adding an `--smv-jars` option that's included in the `--jars` for the `spark-submit`?

AliTajeldin:
So I guess we have two options:
* add `--smv-jars` options and we append those to the generated `--jars` that we pass to `spark-submit` (int addition to the `APP_JAR`)
* parse the spark `--jars` options being passed to `smv-pyrun` and build a new `--jars` that includes the `APP_JAR`

The first option is easier to implement but requires additional documentation.  The second option is harder to implement but is more natural for the user (that is what they would expect).  Since users generally don't read the documentation and will probably expect to be able to supply their own `--jars` option to `smv-pyrun` as part of the spark options, I think we should do option 2 unless it is extremely difficult (shouldn't be).

laneb:
Makes sense.

laneb:
Done.

PR #670, opened 2017-04-25
Support for extra jars using spark --jars option

laneb:
Addressing #627.

### DIFF
--- a/tools/_pyenv.sh
+++ b/tools/_pyenv.sh
@@ -12,5 +12,5 @@ export PYTHONPATH="$SMV_HOME/src/main/python:$PYTHONPATH"
 export SPARK_PRINT_LAUNCH_COMMAND=1
 
 function run_pyspark_with () {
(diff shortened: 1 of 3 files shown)


tools/_pyenv.sh
```diff
@@ -12,5 +12,5 @@ export PYTHONPATH="$SMV_HOME/src/main/python:$PYTHONPATH"
 export SPARK_PRINT_LAUNCH_COMMAND=1
 
 function run_pyspark_with () {
-  spark-submit "${SPARK_ARGS[@]}" --jars "$APP_JAR" --driver-class-path "$APP_JAR" $1 "${SMV_ARGS[@]}"
+  spark-submit "${SPARK_ARGS[@]}" --jars "$APP_JAR,$EXTRA_JARS" --driver-class-path "$APP_JAR" $1 "${SMV_ARGS[@]}"
```
AliTajeldin: What would happen if `EXTRA_JARS` is blank?  Does spark `--jars` option handle a trailing `,`?

laneb: Yes, if EXTRA_JARS is empty there will be a trailing comma (you can see it in the spark submit), but Spark handles this scenario and still includes the APP_JAR.

Issue #643, opened 2017-04-18
manually deleting output data does not trigger re-run but results in exception


XiaosenChen:
In the scala interface, if user manually delete a specific output data (both the .csv dir and .schema dir) of a module, df() on the module will trigger a re-run. However, with python interface, without exiting out of the jupyter, user gets this error.
```
An error occurred while calling z:org.tresamigos.smv.python.SmvPythonHelper.peekStr.
: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 22.0 failed 1 times, most recent failure: Lost task 0.0 in stage 22.0 (TID 33, localhost): java.io.FileNotFoundException: File file:outputdir/outputdata.csv/part-00000 does not exist
```

laneb:
Does this happen every time? If not, can you consistently reproduce it?

laneb:
This can be replicated in `smv-pyshell`. It appears that we are reusing the cached `DataFrame`, which wants to read from the persisted results even after they have been deleted. It is not clear to me how this would have worked before in the Scala shell _or_ the python shell without `ddf`, as we have always cached the `DataFrame`. In any case, to support this behavior `SmvDataSet` will need to skip the cache if there is no persisted output.

laneb:
However, I'm not sure that this is a usage pattern we should encourage. Perhaps it would be better to provide an optional flag in `df` that forces the module to be run again by skipping the cache and ignoring the persisted data.

AliTajeldin:
@laneb 👍 on the `df` flag.  This is related to #481 to ignore persisted files and force a rerun.

laneb:
I will add the flag. What should the behavior be when the user runs a module again in the same shell after deleting its output?

laneb:
Actually, the interface might get a little messy if we also want the flag described in #651. How would we combine the two?

AliTajeldin:
Make both flags optional:
```python
def df(fqn, forceRun=false, version=None):
  ...
```
Then user can provide either positional or keyword mapping during call:
```python
> df("a.b.c", version="v15")
> df("a.b.c", true)
> df("a.b.c", forceRun=true)
```

Another option is to **only** accept keyword args.
```python
def df(fqn, **kw):
  # check here for "forceRun" and "version" args.
```
so the user will have to specify the `forceRun` and `version`

laneb:
Will `df` be documented with `sphinx`? If so I think think the first option is preferable because it provides a more helpful signature.

[...]

Issue #481, opened 2017-02-17
Add smv flag to ignore persisted intermediary files.


AliTajeldin:
Need a `--force-run` or `--ignore-cache` flag to SMV that will tell SMV to ignore the persisted cache files and force the run of all modules even if a persisted file exists.

The current workaround is to delete the output data directory.

This should be implemented after #469

laneb:
Done.

AliTajeldin:
Need to add the command line flags to the user docs.

laneb:
Done.

PR #674, opened 2017-04-25
add optional flag for df to force module to rerun (#643)

laneb:
Addressing (#643). The majority of this commit is adding parentheses to calls to `DataFrame.rdd` - you can't leave of parentheses if there is a default parameter. It should be a short leap from here to #481.

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SmvApp.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvApp.scala
@@ -66,7 +66,7 @@ class SmvApp(private val cmdLineArgs: Seq[String],
    **/
   def createDF(schemaStr: String, data: String = null, isPersistValidateResult: Boolean = false) = {
     val smvCF = SmvCsvStringData(schemaStr, data, isPersistValidateResult)
-    smvCF.rdd
+    smvCF.rdd()
   }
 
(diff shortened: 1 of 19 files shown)


src/main/scala/org/tresamigos/smv/SmvApp.scala
```diff
@@ -98,15 +98,8 @@ class SmvApp(private val cmdLineArgs: Seq[String],
...
-  /**
-   * delete the current output files of the modules to run (and not all the intermediate modules).
-   */
-  private def deleteOutputModules() = {
-    // TODO: replace with df.write.mode(Overwrite) once we move to spark 1.4
-    modulesToRun foreach { m =>
-      m.deleteOutputs()
-    }
-  }
+  private[smv] def deletePersistedResults(dsList: Seq[SmvDataSet]) =
```
AliTajeldin: please add comment that describes this function.

laneb:
For clarity, I will rename `force` argument (of `df`, `runModule`, etc.) to `force_rerun`.

AliTajeldin:
We may not be re-running (the first time this is called). Should probably be `force_run` to be consistent with the `force-run-all` on command line.

laneb:
Yeah `force_rerun` was a typo.

PR #679, opened 2017-04-26
add option --force-run-all to force modules to rerun (#481)

laneb:
Addressing #481.

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SmvApp.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvApp.scala
@@ -22,6 +22,8 @@ import org.apache.spark.{SparkContext, SparkConf}
 import scala.collection.mutable
 import scala.util.{Try, Success, Failure}
 
+import scala.language.postfixOps
+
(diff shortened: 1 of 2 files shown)


src/main/scala/org/tresamigos/smv/SmvApp.scala
```diff
@@ -22,6 +22,8 @@ import org.apache.spark.{SparkContext, SparkConf}
 import scala.collection.mutable
 import scala.util.{Try, Success, Failure}
 
+import scala.language.postfixOps
```
AliTajeldin: we should probably avoid postFixOps in general purpose code.  Only really useful for DSL applications.  It has some other nasty side effects with end of statement inference (hence why you have to import a lib to use it).

laneb: Gotcha, I'll fix that.

laneb: Done.

PR #689, opened 2017-04-28
I381 dependencies stages

Khorkhe:
when running a module whose module dependency is in an undeclared stage, throw error asking user to update the config's stage list.

tests needed to be updated as well as those changes caused a couple of tests to fail.


src/main/scala/org/tresamigos/smv/DataSetResolver.scala
```diff
@@ -38,6 +38,12 @@ class DataSetResolver(val repos: Seq[DataSetRepo],
    */
   def loadDataSet(urns: URN*): Seq[SmvDataSet] =
     urns map { urn =>
+
+      val found = smvConfig.stageNames.map( stage => urn.fqn.startsWith(stage) ).foldLeft(false)(_ || _)
```
laneb: It would be clearer and more succinct to use `exists` here like

`smvConfig.stageNames.exists( stage => urn.fqn.startsWith(stage) )

src/test/python/testSmvFramework.py
```diff
@@ -102,7 +106,7 @@ class SmvRunConfigTest1(SmvBaseTest):
 
     @classmethod
     def smvAppInitArgs(cls):
-        return ['--smv-props', 'smv.config.s=s1',
+        return ['--smv-props', 'smv.config.s=s1', 'smv.stages=testSmvFramework.D4',
```
laneb: A valid stage must be the name of a directory. It cannot be the name of a class, or even the name of a file. Because we assume internally that stages obey these rules, defining stages this way may lead to unexpected results. It would be better to put resources like modules used in this test in a resource directory which can be validly declared a stage.

laneb:
@Khorkhe that sounds like a weird bug. There's no reason why that should cause a problem. By any chance do you have the error handy?

laneb:
Btw @AliTajeldin this PR had another build failure due to a lost executor. Seems like 1/3 builds or so has that problem.

src/test/python/testDataFrameHelper.py
```diff
@@ -31,6 +31,10 @@ def csvAttr(self):
         return self.defaultCsvWithHeader()
 
 class DfHelperTest(SmvBaseTest):
+    @classmethod
+    def smvAppInitArgs(cls):
+        return ['--smv-props', 'smv.stages=testDataFrameHelper.T']
```
Khorkhe: @laneb here's an example of rejected stage name, with the errors that ensue.

Removing `.T` from the specified stage causes 9 test errors when running the python test suite.

Here is one of the 9 error messages:
```
======================================================================
ERROR: test_smvHashSample_with_column (testDataFrameHelper.DfHelperTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/khorkhe/Desktop/twine/SMV/src/test/python/testDataFrameHelper.py", line 49, in test_smvHashSample_with_column
    r1 = df.unionAll(df).smvHashSample(col('k'), 0.3)
TypeError: 'NoneType' object is not callable
```

The remaining 8 errors also show the same error.

However, when setting the stage name to `testSmvFramework` (see code below), the error is different and also misleading:

```
======================================================================
ERROR: test_SmvDQM (testSmvFramework.SmvFrameworkTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/khorkhe/Desktop/twine/SMV/src/test/python/testSmvFramework.py", line 92, in test_SmvDQM
    df.smvDumpDF()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/case.py", line 117, in __exit__
    if not issubclass(exc_type, self.expected):
TypeError: issubclass() arg 2 must be a class or tuple of classes
```

laneb: Cool I'll take a closer look at this later.

laneb:
I don't think the problem is actually caused by the stage not having a dot. More likely it's related to at least one of 

- The stage module (which is being deleted and reloaded) is also the test module
- The stage module is a file (valid stages should be directories)

AliTajeldin:
Note: before this PR can be merged to master, I want us to fully understand why `testDataFrameHelper.T` is actually working (it seems like it shouldn't), and why the name without the `.T` postfix was not working.

src/main/scala/org/tresamigos/smv/DataSetResolver.scala
laneb: Good point - and there already is an `SmvDataSet` method called `parentStage` that does just that.

src/test/python/testDataFrameHelper.py
laneb: This is definitely a weird one. The "'NoneType' object" is the `col` method which is imported from pyspark. There is a test that is calling `runModule`, and right after `runModule` `col` is no longer defined. I don't think it's worth debugging this further right now. Creating the resource directories I described should fix all of these problems.

laneb:
The reason why the stage `testDataFrameHelper.T` is working is 

1. We are only checking at the module `testDataFrameHelper.T` begins with `testDataFrameHelper.T`, not `testDataFrameHelper.T.`
2. When we run `testDataFrameHelper.T` (only happens in `test_smvExportCsv`), we run it without name inference so we don't run into an error trying to import `testDataFrameHelper.T`

src/main/scala/org/tresamigos/smv/DataSetResolver.scala
laneb: The approach that @AliTajeldin will fix a second issue - this implementation only checks that a the fqn starts with `stage`, when it should check that it starts with `stage + "."`. This gives a false positive if there is a single stage `stage1` and the fqn is `stage123`.

Khorkhe:
@laneb would you please clarify what are test resources, and resource-specific directories?

laneb:
A test that requires its own datasets should not define the datasets in the test file. There should be a separate directory for that test's datasets. That way, the test can (among other things) assign the datasets a valid stage. `testHive` gives a good example of this.

Tangentially, it would be a bit clearer if we put all of these resource directories in a directory called `resource` and organized them more explicitly by the name of the test.

Khorkhe:
thanks, will check

laneb:
I will create a doc explaining the conclusion @AliTajeldin, @Khorkhe, and I reached on why the stage name `testDataFrameHelper` was causing problems.

Khorkhe:
ready for review of change requests

PR #694, opened 2017-04-28
I692 edd internal interface

ninjapapa:
Made all _smvEdd, _smv*Hist internal methods to return `EddResultFunctions`, so that Python side can either print the string or use the Edd DF directly.


laneb:
Another lost executor failure from Travis.

PR #695, opened 2017-04-28
I693 apilinks in doc

Khorkhe:
a symlink `latest` is created under scaladocs/ pointing to the latest deployed doc version.

e.g.:
```
$ admin/publish_docs.sh 1.5.2.6 1.5.2.7

// creates a symlink under scaladocs, latest -> 1.5.2.7
// creates a symlink under pythondocs, latest -> 1.5.2.7
```

This way, the URLs to docs in `master` are always pointing to xyzdocs/latest, instead of having to update the URLs in the MD and code files


AliTajeldin:
Actually, I think we should cancel this PR.  The links in the docs should have a version number so when a user visits a version specific doc, they will have the right version pointed.  We modify the link versions during an smv release.  Let's discuss on Monday.

Issue #121, opened 2015-03-27
Add SmvModule level meta data


ninjapapa:
When an SmvModule get persisted, we will calculate meta data in the persisting process. For now, the meta data should have:
- Total number of records
- Real runtime
- EDD dataQATasks (implemented as pipe operation) 

Meta data should be convert as RDD of JSON string and persist with the data file and the same version number.

ninjapapa:
Top level the meta data should be organized as
```javascript
{Name:_,
 Version:_,
 VersionSum:_,
 Log: {...},
 BaseEdd: {...},
 ExtEdd: {...}
}
``` 
where, Name, Version, VersionSum and Log will always be calculated for every SmvModule when persisting. BaseEdd and ExtEdd will be specified through SmvModule's verify method. By default the verify method will create BaseEdd, and will be triggered at the persist step also. Additional verification could be specified with each SmvModule's verify method such as ExtEdd and/or DQM on BaseEdd or ExtEdd.

Log info should also be printed out at run time. 

The following should be included in the Log info:
* Execution time
* Number of records persisted
* DQM reject and/or fixed log

ninjapapa:
Spark 1.2 and after have toJSON method on SchemaRDD,the best way to persist the Meta data could be using SchemaRDD after we migrate to 1.3. 

However, the same persisted Meta data could be reported on the interactive shell or from web interface. Currently EDD already have the shell reporting function coded. The issues is how should we avoid maintaining 2 code bases on reporting on the same data, one in Scala for shell reporting, and on in Javascript for Web reporting.

ninjapapa:
Need a design doc first.

AliTajeldin:
Since we will have quite a few potential meta data objects to track, we should have a global meta config with granular on/off flags for the contents of the meta file.  For example:
```yaml
{
  "name": "blah",  # file level data can always be there.
  "version": 1.0,
  "...": "...",
  "trace" : {
     "run-timestamp": "2017/04/15 08:10:30Z" # job run timestamp (not file creation)
     "inputs": ["file1_abc123.csv", "file2_aa14c3.csv", ...]
  }
  "edd" : {...}
}
```

In the config (app, user, or command line), we can turn on/off certain part of the generated meta data.  Anything that requires extra computation should have an associated flag.  For example:
```java
smv.meta = true; // global flag.  If off, the meta file is not generated.
smv.meta.edd = bool; // determine if edd report should be generated.
[...]

PR #701, opened 2017-05-01
I121 smv meta data

laneb:
Addressing #121.


src/main/scala/org/tresamigos/smv/SmvMetaData.scala
```diff
@@ -0,0 +1,86 @@
...
+      }
+      .toArray
+
+  def toSparkMetadata: Metadata =
+    builder.build
+
+  /**
+   * String representation is a minified json string
+   */
+  override def toString =
```
AliTajeldin: * rename to `toJson`
* combine with toSparkMetadata.

Issue #693, opened 2017-04-28
dead docs links


Khorkhe:
[SMV functions](https://github.com/TresAmigosSD/SMV/blob/master/docs/user/functions.md) page has a couple of links to other doc pages such as DFHelpers, ColumnHelpers etc., all of which are 404 not found.

this is probably high priority.

Khorkhe:
fixed

PR #708, opened 2017-05-03
#693: update doc links with version number

Khorkhe:
cancelled previous pull request. since we currently have perl script handle the update of smv versions in all MD files, simply added version number to the dead docs links


AliTajeldin:
LGTM. merging.

Issue #711, opened 2017-05-04
function to get the stage of a module


Khorkhe:
there is need for a function that checks whether a urn is defined in a Set of stageNames.
implementation would be something like:
```
smvConfig.stageNames.exists( stage => urn.fqn.startsWith(stage + ".") )
```

Currently, a similar function exists in SmvDataSet:
```
private[smv] lazy val parentStage: Option[String] = app.dsm.stageForUrn(urn)
```

However we need to be able to pass the urn as parameter (basically, like app.dsm.stageForUrn) , so the function can be called in DataSetResolver.

Here are some ways of achieving this:

 - make SmvApp available in DataSetResolver, this way we can use the datasetmgr function directly.  Probably not the best idea since it would involve having SmvApp where it is otherwise unneeded.

 - @laneb suggested to define this function in the URN class

None of these seem optimal to me. @AliTajeldin any preference (or alternative)?

Khorkhe:
the function in DataSetMgr was entirely replaced by an equivalent function in URN, as well as all its usages.

Khorkhe:
the current signature for this function is:
```
def getStage(allStageNames: Seq[String]): Option[String] = ...
```
eventually, we will need to remove the stageNames parameter and get them directly from the app, inside the function.

this will have to wait until #675 is completed

AliTajeldin:
Why is this ticket dependent on #675?  This is a one liner that just gets the stage names from config and iterates through them.

laneb:
This would require that there is a configured SmvApp available. The
DependencyRule test needs this function but does not currently extend
SmvTestUtil, so does not have access to SmvApp. I significantly rewrote
this test to use SmvApp on the branch for #675, so we shouldn't duplicate
the work here. I will get that branch ready for a PR shortly.

laneb:
The branch #675 is merged in to master, so this can move ahead.

Khorkhe:
great thanks

Khorkhe:
fixed

PR #717, opened 2017-05-05
#711: function to get stage of urn, to replace the one in datasetmgr

Khorkhe:


Khorkhe:
a first pass for the function implementation.  Additional changes will need to be made later on, see comments on #711

Issue #675, opened 2017-04-26
Remove DependecyRules class


laneb:
Currently there are certain requirements regarding the dependencies of `SmvModule` which are defined and enforced using a class called `DependencyRules`. This abstraction is not helpful and has to be propagated from `SmvApp` down to the `DataSetResolver`. We should remove this class and implement the logic of the rules in `DataSetResolver`.

laneb:
Depends on #381, because we need to be sure that all modules have stages to check dependency rules.

PR #718, opened 2017-05-05
I675 rm dependency rules

laneb:
Addressing #675.


laneb:
Made this prematurely. Will reopen when the branch is ready.

Issue #696, opened 2017-04-29
add/refactor get, create, update dataset functionality in smvserver


AntonioRecaldeRusso:
- We should be able to get dataset properties based on its stage name, basename, or fqn
- The server should allow the creation of new datasets based on a new fqn, description, dsType, isEphemeral, setting the remaining properties (e.g.: run method, requiresDS (modules), tableName (hive), path (csv)) set to default values
- The server should update a dataset (e.g.: update body of run method) based on a post request with containing the new values

PR #719, opened 2017-05-06
Get create update dataset  #696

AntonioRecaldeRusso:
- Adds functionality to update full source code of dataset,
- fixes some errors in method body detection,
- Fixes read of requiresDS when requiring smv links (server returns fqn of link target)
- get app info functionality


src/main/python/smvserver.py
```diff
@@ -585,7 +625,7 @@ def updateModuleMetaData():
         module = DataSetRepoFactory(SmvApp.getInstance()).createRepo().loadDataSet(fqn)
     except ImportError:
         # dataset does not exist.. create new module
-        stagesList = ['stage1', 'stage2'] # TODO: get actual list of stages.. not hardcoded
+        stagesList = ['airlineapp.etl', 'airlineapp.datamodel', 'airlineapp.feature']  # TODO: get actual list of stages.. not hardcoded
```
laneb: Should this todo be completed before merge?

AliTajeldin: @laneb it's ok. It is a work in progress.

PR #721, opened 2017-05-08
I675 rm dependency rules

laneb:
Addressing #675. Includes rewriting the `DependencyRule` test to check that loading datasets succeeds and fails in all the same scenarios, but without `DependencyRule` itself.


src/main/scala/org/tresamigos/smv/urn.scala
```diff
@@ -24,6 +24,8 @@ package org.tresamigos.smv
 sealed abstract class URN(prefix: String) {
   def fqn: String
   override def toString: String = s"${prefix}:${fqn}"
+  def toModURN: ModURN          = ModURN(fqn)
```
AliTajeldin: in situations like these, I really hate scala :-)

Issue #739, opened 2017-05-10
extract functions from smv-server module


AliTajeldin:
The current entry points all have the implementation embedded with the entry point definition.  Should extract the implementation out and the entry points should only be concerned with extracting request parameters and building the response. 

Note: needs to be done **after** removal of "update modules" entry points.

ninjapapa:
done.

PR #741, opened 2017-05-10
I738 server port

AliTajeldin:


laneb:
lgtm

jacobdr:
That `src/main/python/smvserver.py` file is getting to be pretty messy....

AliTajeldin:
@jacobdr: Agree, see #739

Issue #742, opened 2017-05-10
Clean up sphinx-conf.py


laneb:
There is a lot of boilerplate code in there (like configuration for latex and epub documents) that is irrelevant to us. Also we should update it to include the SMV version and author "TresAmigosSD" (right now the author is "Author".

AliTajeldin:
done

PR #750, opened 2017-05-11
I742 clean sphinx conf

laneb:
Addressing #742. Although sphinx-conf.py now has the SMV release version, we will not need to update it because it will track with .smv_version.


admin/conf/sphinx-conf.py
```diff
@@ -48,18 +48,20 @@
...
 #
-# The short X.Y version.
-version = u''
 # The full version, including alpha/beta/rc tags.
-release = u''
+# The location of this file is not reliable so we cannot read .smv_version directly
+# Instead the version is exported from the script that creates the sphinx build
+release = os.environ["SMV_VERSION"]
+# The short X.Y version.
+version = release[:2]
```
AliTajeldin: `release[:2]` would be `1.` for version `1.5.2.2`.  Did you mean `:3` instead?

laneb: Yes I did! Also I wasn't able to find anywhere where the "version" is actually used by Sphinx, but it's part of their default configuration.

laneb: Fixed.

Issue #651, opened 2017-04-19
SmvModuleLink should be able to be resolved in shell


ninjapapa:
Here is the use case:
* Published module A in stage1 with version 123
* In stage2 we have a link `ALink` defined as `ALink = SmvModuleLink(stage1.A)`, and
* The config file specified the version of stage1 is `123`
With above setup, we want to check/profile on the published `A`. 

Ideally in the pyshell or jupyter we should be able to:
```
a = df("ALink") 
a.smvEdd()
```

Currently to do above, we need to create a dummy SmvModule to pass down the link.

AliTajeldin:
How about adding an optional version string to `df()`.  That is something like:
```
a = df("A", "123")
b = df("A", "124")
c = df("A")
```
That would also allow you to compare different published versions against the current code version.

jacobdr:
Really like this feature idea. I imagine we could then even use DQM to raise issue if the diff is unexpectedly large.

Khorkhe:
fixed, as per Ali's suggestion above

ninjapapa:
Need to update docstr and user doc for the new parameter of "df".

Khorkhe:
fixed

PR #761, opened 2017-05-14
#651: let DF take an optional version number which, if specified, tri…

Khorkhe:
…es to read published data for that version

### DIFF
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -146,14 +146,14 @@ class SmvApp(object):
         """
         return self.j_smvApp.generateAllGraphJSON()
 
-    def runModule(self, urn, forceRun = False):
+    def runModule(self, urn, forceRun = False, version = None):
         """Runs either a Scala or a Python SmvModule by its Fully Qualified Name(fqn)
         """
-        jdf = self.j_smvPyClient.runModule(urn, forceRun)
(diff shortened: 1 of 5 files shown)


Khorkhe:
if we add a check that only allows to specify either the forceRun or the publish version, users would have to resort to use named params.  Is this desirable? alternatively, whatever forceRun flag they specify can be ignored.

src/main/python/smv/smvapp.py
```diff
@@ -146,14 +146,14 @@ def get_graph_json(self):
...
         """Runs either a Scala or a Python SmvModule by its Fully Qualified Name(fqn)
         """
-        jdf = self.j_smvPyClient.runModule(urn, forceRun)
+        jdf = self.j_smvPyClient.runModule(urn, forceRun, self.scalaOption(version))
         return DataFrame(jdf, self.sqlContext)
 
-    def runModuleByName(self, name, forceRun = False):
-        jdf = self.j_smvApp.runModuleByName(name, forceRun)
+    def runModuleByName(self, name, forceRun = False, version = None):
+        jdf = self.j_smvApp.runModuleByName(name, forceRun, self.scalaOption(version))
```
laneb: I was curious why this works because I think `None` in Python usually corresponds to `null` in Scala. Apparent `Option(null)` results in `None`. Cute.

Khorkhe: Interesting

Issue #700, opened 2017-05-01
Need a python conf/smv_shell_init.py 


ninjapapa:
Currently on scala side, when run `smv-shell`, a project level script `conf/shell_init.scala` will be loaded at the beginning of the SparkShell. We need an equivalent on the python side. Basically each project could have a `conf/smv_shell_init.py` file, and `smv-pyshell` and `smv-jupyter` should import it if exist.

laneb:
Should it be imported or executed?

ninjapapa:
python `import` is executing, right?

laneb:
Importing `smv_shell_init` will execute the code, but in the context of the module `smv_shell_init`, so the user will have to call their function `foo` like `smv_shell_init.foo`. To execute client startup code, pyspark uses the builtin function `execfile`, with an effect equivalent to pasting the file into the shell when it starts up. Actually, we might be able to do `from smv_shell_init import *`, but I'm not sure all of the differences in behavior there might be. It _would_ make it easy for users to ignore some of the attributes defined in the module, like helper functions. It _wouldn't_ allow users to use any of the SMV shell methods like `df`

laneb:
Notably, `pyspark` uses `execfile` to run the user's `PYTHON_STARTUP`. However, based on [this](http://stackoverflow.com/questions/11703327/python-import-vs-execfile) and [this](http://stackoverflow.com/questions/27517003/difference-between-import-and-execfile) SO posts, it sounds like `import` is probably better and also preferred even for this kind of use case.

PR #771, opened 2017-05-16
i700 smv-pyshell user init

laneb:
Addresses #700.

### DIFF
--- a/tools/conf/smv_pyshell_init.py.template
+++ b/tools/conf/smv_pyshell_init.py.template
@@ -52,6 +52,14 @@ sc.setLogLevel("ERROR")
 
 user_args = "_SMV_ALL_ARGS_".split()
 all_args = ['--cbs-port', str(callback_server_port)] + user_args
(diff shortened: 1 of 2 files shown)


tools/conf/smv_pyshell_init.py.template
```diff
@@ -52,6 +52,14 @@ sc.setLogLevel("ERROR")
...
+app = SmvApp.createInstance(all_args, sc, sqlContext)
 
 from smv.smvshell import *
+
+# Import user-defined helpers
+app.prepend_source("conf/")
+try:
+  from smv_shell_init import *
+# File may not exist
+except ImportError:
```
AliTajeldin: What happens if there are errors in `smv_shell_init.py`?
Rather than ignore import errors we should check if the file exists and only import it if it exists (without try/catch block)

laneb: Fixed.

PR #779, opened 2017-05-17
I764 smv export csv on cluster

ninjapapa:


AliTajeldin:
@jacobdr : I think the issue is that the main entry point (`smvExportCsv`) is a DF helper and not at the `smvDataset` level.  So there is no way for us to determine if there is already a saved version on disk.  The good new is if that was already persisted, then the DAG for the DF is just a read.  So we would just be doing a straight copy on HDFS.
We may consider adding an `smvExportCSV` method to `SmvDataset` that can be a bit more efficient but that is a separate ticket I think.

Issue #770, opened 2017-05-16
Enable use of links when updating requiresDS of a dataset via the smv server


AntonioRecaldeRusso:
If a required dataset is not in the same stage as the dataset being updated, the required dataset should be wrapped around a SmvPyModuleLink function

laneb:
Is this issue complete?

AntonioRecaldeRusso:
completed. Can be closed

PR #785, opened 2017-05-19
#770: datasets in a different stage are turned into Smv links in requiresDS (smvserver.py)

AntonioRecaldeRusso:
Will turn required dataset into SmvModuleLink if not in same stage as dataset being updated.

### DIFF
--- a/src/main/python/scripts/smvserver.py
+++ b/src/main/python/scripts/smvserver.py
@@ -438,7 +438,7 @@ def extractImportsFromFqns(fqns = []):
     '''
     return list(map(lambda fqn: ".".join(fqn.split(".")[:-1]), fqns))
 
-def buildImports(className = None, requiresDS = None, linkFqnList = []):
+def buildImports(className = None, requiresDS = None, requiredDSLinks = []):
     '''Template for generating the imports section of a dataset'''
     # classname not provided, then create global imports
     if not className:
@@ -453,7 +453,7 @@ from pyspark.sql.functions import *
 
     dependencies = extractImportsFromFqns(requiresDS)
     importStatements = "\n".join(map(lambda dep: "import {}".format(dep), dependencies))
-    linkDefinitions = "\n".join(["{0} = SmvPyModuleLink({1})".format(l["linkName"], l["fqn"]) for l in linkFqnList])
+    linkDefinitions = "\n".join(["{0} = SmvModuleLink({1})".format(l.varName, l.fqn) for l in requiredDSLinks])
     return """\
 ###---{0}_IMPORTS_START---###
 {1}
@@ -513,18 +513,15 @@ def getRequiredModulesWithLinks(currentFqn, fqns):
             requiresDSList.append(fqn)
         # if different stage, wrap around link
         else:
-            # link name
-            linkName = fqn.replace(".", "_") + "_link"
-            requiresDSList.append(linkName)
+            requiresDSList.append(RequiredDSLink(fqn).varName)
     return requiresDSList
 
-def getRequiredDSLinkVars(currentFqn, fqns):
+def getRequiredDSLinks(currentFqn, fqns):
     linkVarsList = []
     currentStage = getStageFromFqn(currentFqn)
     for fqn in fqns:
         if currentStage != getStageFromFqn(fqn):
-            linkName = fqn.replace(".", "_") + "_link"
-            linkVarsList.append({ "linkName": linkName, "fqn": fqn})
+            linkVarsList.append(RequiredDSLink(fqn))
     return linkVarsList
 
 # csv
@@ -588,13 +585,6 @@ JOB_SUCCESS = 'SUCCESS: Code updated!' # TODO: rename CODE_UPDATE_SUCCESS
 OK = ""
 
 MODULE_DOES_NOT_EXIST= 'MODULE_DOES_NOT_EXIST'
-
-@app.route("/api/test", methods = ['POST'])
-def test():
-    currentFqn = request.form["fqn"].encode("utf-8")
-    fqns = ast.literal_eval(request.form["requiresDS"])
-    return jsonify(getRequiredDSLinkVars(currentFqn, fqns))
-
 
 @app.route("/api/get_app_info", methods = ['POST'])
 def get_app_info():
@@ -891,7 +881,7 @@ def updateDatasetInfo():
                 (classImportsStart, classImportsEnd) = getPlutoImportsStartEnd(lines_of_code_list, className)
                 # process requiresDS, turn them into links, get link definitions
                 reqDSWLinks = getRequiredModulesWithLinks(fqn, requiresDS)
-                reqDSLinkVars = getRequiredDSLinkVars(fqn, requiresDS)
+                reqDSLinkVars = getRequiredDSLinks(fqn, requiresDS)
                 classImportsSection = buildImports(className, requiresDS, reqDSLinkVars).splitlines(True)
                 if classImportsStart:
                     newFileContents = classImportsSection + newFileContents[:classImportsStart] + newFileContents[classImportsEnd + 1:]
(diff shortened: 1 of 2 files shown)


src/main/python/scripts/smvserver.py
```diff
@@ -367,10 +367,21 @@ def get_module_name_from_fqn(fqn):
 def getMsnFromFqn(fqn):
     '''based on a fqn, returns an object with the basename, and stage for that given fqn'''
     fqn_split = fqn.split(".")
-    stage = ".".join(fqn_split[:-2])
+    stage = ".".join(fqn_split[:-2])  # TODO: fix.. if stage a.b and fqn a.b.c.d.e.f, will return a wrong stage
```
AliTajeldin: Please fix.

src/main/python/scripts/smvserver.py
```diff
@@ -367,10 +367,21 @@ def get_module_name_from_fqn(fqn):
...
     baseName = fqn_split[-1]
     return { "stage": stage, "baseName": baseName }
 
+def getStageFromFqn(fqn):
+    '''returns the stage given a a dataset's fqn'''
+    # constructing urn for dataset
+    try:
+        urn = DataSetRepoFactory(SmvApp.getInstance()).createRepo().loadDataSet(fqn).urn()
```
AliTajeldin: DRY!

src/main/python/scripts/smvserver.py
```diff
@@ -367,10 +367,21 @@ def get_module_name_from_fqn(fqn):
...
+        # getting stage from URN factory in scala side
+        stage = SmvApp.getInstance()._jvm.org.tresamigos.smv.URN.apply(urn).getStage().get().encode("utf-8")
```
AliTajeldin: too much internal knowledge outside the app. `_jvm.org.tresamigos.smv.URN.apply(urn).getState().get().encode` should be in SmvApp python side.

src/main/python/scripts/smvserver.py
```diff
@@ -444,10 +455,14 @@ def buildImports(className = None, requiresDS = None):
 
     dependencies = extractImportsFromFqns(requiresDS)
     importStatements = "\n".join(map(lambda dep: "import {}".format(dep), dependencies))
+    linkDefinitions = "" if not linkFqnList \
+        else "\n".join(map(lambda l: "{0} = SmvPyModuleLink({1})".format(l["linkName"], l["fqn"]), linkFqnList))
```
AliTajeldin: * use list comprehension here. `[f(l) for l in linkFqnList]`
* use empty list as default for `linkFqnList` instead of None.  No need to check.

src/main/python/scripts/smvserver.py
```diff
@@ -488,6 +503,30 @@ def buildRequiresDS(requiresDS, indentation="\t"):
...
+    currentStage = getStageFromFqn(currentFqn)
+
+    for fqn in fqns:
+        # if same stage, fqn
+        if currentStage == getStageFromFqn(fqn):
+            requiresDSList.append(fqn)
+        # if different stage, wrap around link
+        else:
+            # link name
+            linkName = fqn.replace(".", "_") + "_link"
```
AliTajeldin: This should be creating an instance of `Link` class.

src/main/python/scripts/smvserver.py
```diff
@@ -488,6 +503,30 @@ def buildRequiresDS(requiresDS, indentation="\t"):
 {1}def requiresDS(self):\n\
 {1}{1}return [{0}]\n".format(requires, indentation)
 
+def getRequiredModulesWithLinks(currentFqn, fqns):
```
AliTajeldin: separate out into two functions.  one for linkVars and one for requires.

src/main/python/scripts/smvserver.py
laneb: Also, it isn't very to create a repo outside of the `DataSetMgr`. Implicitly this creates a new transaction and invalidates modules loaded in previous transactions but we should generally go through the `DataSetMgr` when we create transactions. My suggestions are

1. If all you need to accomplish is to get the URN for the module, then just use the `ModURN` constructor

2. If you are trying to force this to fail when the module doesn't exist, load it through the `DataSetMgr`. Then you can also the actual `URN` object from the module it gives you and skip most of the next line.

src/main/python/scripts/smvserver.py
laneb: +1 for list comprehensions 😸

laneb: Also, I note that you are using `if not x` in circumstances when you are trying to test if some `x` is not `None`. In Python, `if x is not None` is generally preferred. It reads easier IMO and also prevents hiding bugs that could arise if the function was incorrectly called with Falsy values like an empty list.

src/main/python/smv/smvapp.py
```diff
@@ -160,6 +160,14 @@ def urn2fqn(self, urnOrFqn):
         """Extracts the SMV module FQN portion from its URN; if it's already an FQN return it unchanged"""
         return self.j_smvPyClient.urn2fqn(urnOrFqn)
 
+    def fqn2urn(self, fqn):
+        """Returns urn string for a given fqn"""
+        return self._jvm.org.tresamigos.smv.ModURN(fqn).toString()
+
+    def getStageFromFqn(self, fqn):
+        """Returns the stage name for a given fqn"""
+        return self._jvm.org.tresamigos.smv.URN.apply(self.fqn2urn(fqn)).getStage().get()
```
laneb: This code is confusing. You are constructing a URN to get its string (`fqn2urn`)
so you can construct a URN _again_ and then get its stage. It would be clearer and  more direct to have a method for getting the `ModURN` of a module and then just use that in both `fqn2urn` and `getStageFromFqn`.

src/main/python/smv/smvapp.py
```diff
@@ -160,6 +160,14 @@ def urn2fqn(self, urnOrFqn):
         """Extracts the SMV module FQN portion from its URN; if it's already an FQN return it unchanged"""
         return self.j_smvPyClient.urn2fqn(urnOrFqn)
 
+    def fqn2urn(self, fqn):
+        """Returns urn string for a given fqn"""
```
laneb: You should note in the docstring that this will be a module URN and not a link URN.

src/main/python/scripts/smvserver.py
AntonioRecaldeRusso: #6b81f53

src/main/python/scripts/smvserver.py
AntonioRecaldeRusso: Added getStageFromFqn method to smvapp in #6b81f53

src/main/python/scripts/smvserver.py
AntonioRecaldeRusso: Added fqn2urn method to smvapp in #6b81f53.. Using ModURN as suggested

src/main/python/scripts/smvserver.py
AntonioRecaldeRusso: using list comprehensions #6b81f53

src/main/python/scripts/smvserver.py
AntonioRecaldeRusso: f80ca87

src/main/python/scripts/smvserver.py
AntonioRecaldeRusso: dc8f0a1

src/main/python/smv/smvapp.py
AntonioRecaldeRusso: 5934fca

src/main/python/smv/smvapp.py
AntonioRecaldeRusso: removed in 5934fca

PR #793, opened 2017-05-20
I723 jdbc query

laneb:
Still needs to be tested with connections to different DBs. Plays nicely with MySQL and the tests use Derby.


laneb:
I will add an example of `SmvJdbcTable`. Should I also and an example of `--publish-hive`? There is already info about the cli option and the url property in the correct docs.

laneb:
Nevermind, included an example with `--publish-jdbc`.

PR #796, opened 2017-05-20
I795 description

AntonioRecaldeRusso:
description now updates correct class docstring

### DIFF
--- a/src/main/python/scripts/smvserver.py
+++ b/src/main/python/scripts/smvserver.py
@@ -499,14 +499,23 @@ def buildRequiresDS(requiresDS, indentation="\t"):
 {1}def requiresDS(self):\n\
 {1}{1}return [{0}]\n".format(requires, indentation)
 
-def getRequiredModulesWithLinks(currentFqn, fqns):
-    '''takes a fqn (currentFqn) and a list of fqns or SmvModuleLink(s) required by the currentFqn
-        if a required fqn is in the same stage, it will remain as a fqn in the returned list. If
-        the required fqn is in a different stage, it will be subtituted by a SmvModuleLink'''
+def getRequiredModulesWithLinks(currentFqn, requiredFqns):
+    '''takes as arguments a fqn (currentFqn) and a list of required fqns.
+        Returns a list as it should be consumed by the requiresDS method of a dataset.
+        The returned list contains:
+        - fqn: for every requiredFqn that is in the same stage as currentFqn
+        - SmvModuleLink variable name: for every requiredFqn in a different stage than currentFqn
+         E.g:
+        currentFqn = 'stage1.foo.foo'
+        requiredFqns= [stage1.bar.bar, stage2.baz.baz]
+        Will return: [
+            stage1.foo.foo, # it is in same stage as current fqn, so returns fqn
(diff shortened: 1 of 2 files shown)


src/main/python/scripts/smvserver.py
```diff
@@ -488,6 +499,30 @@ def buildRequiresDS(requiresDS, indentation="\t"):
 {1}def requiresDS(self):\n\
 {1}{1}return [{0}]\n".format(requires, indentation)
 
+def getRequiredModulesWithLinks(currentFqn, fqns):
```
AliTajeldin: rename `fqns` to `requiredFqns`.

src/main/python/scripts/smvserver.py
```diff
@@ -488,6 +499,30 @@ def buildRequiresDS(requiresDS, indentation="\t"):
...
+    '''takes a fqn (currentFqn) and a list of fqns or SmvModuleLink(s) required by the currentFqn
```
AliTajeldin: This comment is a bit confusing.  Are we taking a list that has `SmvModuleLink(s)` or are we creating a list that has `SmvModuleLinks`?

src/main/python/scripts/smvserver.py
```diff
@@ -488,6 +499,30 @@ def buildRequiresDS(requiresDS, indentation="\t"):
...
+    for fqn in fqns:
+        # if same stage, fqn
+        if currentStage == getStageFromFqn(fqn):
+            requiresDSList.append(fqn)
+        # if different stage, wrap around link
+        else:
+            requiresDSList.append(RequiredDSLink(fqn).varName)
+    return requiresDSList
+
+def getRequiredDSLinks(currentFqn, fqns):
```
AliTajeldin: what is the difference between this function and the one above?
Needs a method comment to explain.

AliTajeldin:
Btw, how did a bunch of commit for ticket #770 get added to this PR?

src/main/python/scripts/smvserver.py
```diff
@@ -488,6 +499,30 @@ def buildRequiresDS(requiresDS, indentation="\t"):
...
+    requiresDSList = []
+    currentStage = getStageFromFqn(currentFqn)
+
+    for fqn in fqns:
+        # if same stage, fqn
+        if currentStage == getStageFromFqn(fqn):
+            requiresDSList.append(fqn)
+        # if different stage, wrap around link
+        else:
+            requiresDSList.append(RequiredDSLink(fqn).varName)
```
AliTajeldin: Are we just creating the `RequiredDSLink` object to throw it away? seems weird.

AntonioRecaldeRusso: It is done so to transform the fqn into the SmvModuleLink variable name. Namely, it avoids having to repeat the link `Namely "{}_link".format(fqn.replace(".", "_"))`
It is also possible to create a static method in the RequiredDSLink class that will transform the fqn into the var name, and thus avoid creating an object and throwing it away.

src/main/python/scripts/smvserver.py
AntonioRecaldeRusso: comments added e481809

PR #803, opened 2017-05-22
I726 user schema

laneb:


AliTajeldin:
really nice!

src/main/python/smv/smvdataset.py
```diff
@@ -330,8 +330,16 @@ class SmvCsvFile(WithParser, SmvInput):
...
         self._smvCsvFile = smvApp.j_smvPyClient.smvCsvFile(
-            self.fqn(), self.path(), self.csvAttr(),
-            self.forceParserCheck(), self.failAtParsingError())
+            self.fqn(),
+            self.path(),
+            self.csvAttr(),
+            self.forceParserCheck(),
+            self.failAtParsingError(),
+            smvApp.scalaOption(self.userSchema())
+        )
```
AntonioRecaldeRusso: This change produces runtime errors. Reverting to the old version of this constructor eliminates an error observed when trying to get the dependency graph json.

laneb:
Did you make sure to recompile? The change in this constructor is paired
with a change in the Scala SmvCsvFile constructor.

PR #806, opened 2017-05-22
I559 dry run option

Khorkhe:


src/main/scala/org/tresamigos/smv/SmvApp.scala
```diff
@@ -186,6 +186,25 @@ class SmvApp(private val cmdLineArgs: Seq[String],
...
 
   /**
+   * execute as dry-run if the dry-run flag is specified.
+   * This will show which modules are not yet persisted that need to run, without
+   * actually running the modules.
+   * @return true if dry-run option was specified, otherwise false
+   */
+  private def dryRun(): Boolean = {
+    if (smvConfig.cmdLine.dryRun()) {
+      val modsNotPersisted = modulesToRun.filterNot(mod => mod.isPersisted || mod.isEphemeral)
```
AliTajeldin: This is only showing of the modules the user specified to run which should be run, not the actual modules that need to be run.  For example if A depends on B and B was changed, then the results of `smv-pyrun --dry-run -m A` should show both A and B.  The above code only shows A.

Issue #547, opened 2017-03-18
Make Python datasetHash of SmvCsvFile and SmvMultiCsvFiles depend on the files


ninjapapa:
On Scala side, the datasetHash of SmvCsvFile actually depends on the file's name and last update time. Should support the same on Python side, also need to make SmvMultiCsvFiles' datasetHash depends on the files.

laneb:
Since the Python SmvCsvFile as a reference to its Scala counterpart, we should be able to delegate to the Scala implementation of datasetHash for all of the inputs.

ninjapapa:
I'm considering to separate a "sourceHash" method from the Scala SmvFile implementation, and make it an abstract method of `SmvInput`, while `datasetHash` of `SmvInput` will be `sourceHash + classCrc`.

AliTajeldin:
Not sure I follow the proposed solution above with the "sourceHash" method.
I think we currently create an actual `SmvCsvFile` on the scala side right now.  Might as well use the hash value from the scala side.  It will be a ping-pong of calls but that is only done at init time.

laneb:
I think I follow. `datasetHash` of the Scala `SmvFile` (inherited by the Scala `SmvCsvFile`) combines class CRC with a "sourceHash" which tracks changes in the CSV filename and schema (https://github.com/TresAmigosSD/SMV/blob/master/src/main/scala/org/tresamigos/smv/SmvDataSet.scala#L353). If we wanted to accomplish the same with for Python `SmvCsvFile`, we would need to combine the hash of the Python class's source with the "sourceHash" of the CSV file and schema, which preferrably isn't duplicated on the Python side.

ninjapapa:
@laneb 👆

laneb:
The pattern for `datasetHash` consistently boila down to adding the class CRC to a CRC (if any) of some instance values.  I would propose that that be pattern be formalized so with a `sourceHash` method and an `instanceHash` method (names negotiable). `SmvDataSet` would implement `datasetHash` as the sum of the two. `instanceHash` would normally be 0, except for certain input types for which it would be implemented in Scala. `sourceHash` would be the Scala class CRC, except for `SmvExtModulePython` which would be the Python class CRC.

laneb:
Done.

PR #813, opened 2017-05-24
I547 python csv hash

laneb:
Addressing #547.


src/main/python/smv/smvdataset.py
```diff
@@ -144,7 +143,7 @@ def version(self):
     def isOutput(self):
         return isinstance(self, SmvOutput)
 
-    def datasetHash(self):
+    def sourceCodeHash(self):
```
AliTajeldin: add doc (highlight difference between this and `instanceValHash`.

laneb: Done.

Issue #818, opened 2017-05-25
smv-install script should be tied to SMV version


laneb:
`smv-install` currently expects the SMV version as an argument. Instead the version should be hard-coded in the script. E.g. run `smv-install` from the 1.5.2.9 release will always install SMV 1.5.2.9. This will simplify making changes to the install script, as the latest version of the script will be able to e.g. run new scripts that don't exist in old releases. This version will need to be updated by `smv-release`.

laneb:
Done.

PR #819, opened 2017-05-25
SMV version should be hardcoded in smv-install (#818)

laneb:
Addressing #818.

### DIFF
--- a/tools/smv-install
+++ b/tools/smv-install
@@ -5,9 +5,7 @@ PROG_NAME=$(basename "$0")
 
 function usage()
 {
-  echo "USAGE: ${PROG_NAME} [-spark] smv_version_to_release(n.n.n.n)"
-  LATEST_VERSION=$(curl https://raw.githubusercontent.com/TresAmigosSD/SMV/master/.smv_version)
-  echo "Latest SMV version is: ${LATEST_VERSION:-Could not parse version from Github API}"
+  echo "USAGE: ${PROG_NAME} [-spark]"
   exit $1
 }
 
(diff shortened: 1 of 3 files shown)


tools/smv-install
```diff
@@ -21,7 +19,7 @@ function parse_args()
     DO_INSTALL_SPARK=1
     shift
   fi
-  SMV_VERSION="$1"
+  SMV_VERSION="1.5.2.9"
```
jacobdr: I think that this should be user-configurable if desired. So something like:

`SMV_VERSION="${SMV_VERSION:-1.5.2.9}"`

Previously, I have used the smv-install script from the prior release to download a new version at a different install location on disk, prepended it to my PATH, and then tested the new version against our test suite. This PR in its current form makes it harder to do that I think

laneb: It is true that you would not be able to use e.g. the 1.5.2.10 script to install 1.5.2.11. Can you elaborate on why that would be preferable over following the installation instructions for 1.5.2.11?

AliTajeldin: There is more to the hardcoding of the version in the install script than just not requiring the user to supply the version on the command line.  We shouldn't have to make the script backwards compatible with old versions.
To install older versions, we can just substitute the desired version number in the `curl` command used to download and run the script in the README.  We should document how to install older versions.

AliTajeldin: put comment above this to indicate that it is automatically set by the `smv-release` script.

AliTajeldin:
Merging right now to make smv release.  Add comment to master later.

tools/smv-install
laneb: Done.

PR #829, opened 2017-05-29
I828 smvserver

AntonioRecaldeRusso:
This pr allows extending the smv server by providing  arguments
"--start-server <projects dir> -e <python script name>" when running a smv image container.

### DIFF
--- a/docker/smv/entrypoint.sh
+++ b/docker/smv/entrypoint.sh
@@ -36,11 +36,20 @@ fi
 sudo -u ${USER_NAME} -i bash -c "mkdir -p /projects/.ivy2"
 cd /projects
 
+E_EXTENDED_SMV_SERVER_SCRIPT=''
+
 if [[ $# == 0 ]]; then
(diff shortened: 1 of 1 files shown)


docker/smv/entrypoint.sh
```diff
@@ -36,11 +36,20 @@ fi
...
 
+E_EXTENDED_SMV_SERVER_SCRIPT=''
+
 if [[ $# == 0 ]]; then
     # start bash if user did not supply parameters.
     sudo -u ${USER_NAME} -i bash
 elif [[ $1 == "--start-server" ]]; then
     SERVER_PROJ_DIR="${2:?must provide project directory}"
+
+    if [[ $# -gt 2 ]]; then
```
AliTajeldin: you can skip check for $# here and just check if "$3" is "-e" (note that you must use double quotes around the "$3")

AntonioRecaldeRusso: 278cd35

Issue #827, opened 2017-05-29
smv-server getFqnOfRequire failed


ninjapapa:
Since SmvModuleLink now is a class, and "target" is an attribute instead of a method, `getFqnOfRequire` of smv-server stoped working.

This issue can be easily addressed when making the SmvModuleLink, if smv-server has some test case. This issue is a good example of why we need #826

laneb:
Is this issue complete?

AntonioRecaldeRusso:
completed. Can be closed

PR #831, opened 2017-05-30
#827: getting target from SmvModuleLink now not via a method

AntonioRecaldeRusso:
In order to get the fqn of smv module link target, we were using the SmvModuleLink's .target() method.
.target() is no longer a method, so this code adapts to the given change.


jacobdr:
Thank you!
Spark by Readdle

PR #832, opened 2017-05-30
I479 python error messages

Khorkhe:


src/main/python/smv/stacktrace_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+    """abstract class that can be mixed in to allow functions to use the
+        'with_stacktrace' decorator.
+        This will enable proper error reporting in functions that can be overridden
+        by either end-users or developers
+    """
+    __metaclass__ = abc.ABCMeta
+
+    def __getattribute__(self, name):
+        if name in FUNC_LIST:
+            obj_getattribute = object.__getattribute__(self, name)
```
laneb: This code would read easier if the name was more specific about what we expect the attribute to be (like "user_function" or something).

src/main/python/smv/stacktrace_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+
+            def catche_errs(*pargs, **kwargs):
```
laneb: Nice catch with the **kwargs.

src/main/python/smv/stacktrace_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+                try:
+                    res = obj_getattribute(*pargs, **kwargs)
+                except BaseException as err:
+                    traceback.print_exc()
```
laneb: Can you verify that when the trace is printed we get the correct function? Not a problem if the trace also includes catche_errors, but e.g. if the function being called is isEphemeral then the stacktrace shhould include isEphemeral and the correct line in the user's code.

Khorkhe: When intentionally adding a typo in an smvModule's ephemeral function, for example, the trace looks like:

```
Traceback (most recent call last):
  File "/Users/khorkhe/Desktop/twine/SMV/src/main/python/smv/stacktrace_mixin.py", line 20, in catche_errs
    res = obj_getattribute(*pargs, **kwargs)
  File "/Users/khorkhe/Desktop/temp/enter/src/main/python/stage2/category.py", line 20, in isEphemeral
    return Truee
NameError: global name 'Truee' is not defined
```

It's quite accurate.

src/main/python/smv/stacktrace_mixin.py
laneb: Actually something like "func_needing_stacktrace" maybe.

src/main/python/smv/stacktrace_mixin.py
laneb: niiiiice

src/main/python/smv/stacktrace_mixin.py
Khorkhe: aight, replaced

src/main/python/smv/stacktrace_mixin.py
Khorkhe: :)

src/main/python/smv/stacktrace_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+        This will enable proper error reporting in functions that can be overridden
+        by either end-users or developers
+    """
+    __metaclass__ = abc.ABCMeta
+
+    def __getattribute__(self, name):
+        if name in FUNC_LIST:
+            func_needing_stacktrace = object.__getattribute__(self, name)
+
+            def catche_errs(*pargs, **kwargs):
```
AliTajeldin: typo in function name.

src/main/python/smv/stacktrace_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+                    raise err
+
+                return res
+
+            return catche_errs
+        else:
+            return object.__getattribute__(self, name)
+
+def with_stacktrace(func):
+    FUNC_LIST.append(func.__name__)
```
AliTajeldin: `func.__name__` alone is ambiguous.  need the class name as well.

src/main/python/smv/stacktrace_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+
+class WithStackTrace(object):
+    """abstract class that can be mixed in to allow functions to use the
+        'with_stacktrace' decorator.
+        This will enable proper error reporting in functions that can be overridden
+        by either end-users or developers
+    """
+    __metaclass__ = abc.ABCMeta
+
+    def __getattribute__(self, name):
```
AliTajeldin: given how tricky this was to implement, need some doc on how we are using `__getattribute__` and looking up the names later to wrap the error handling.

src/main/python/smv/stacktrace_mixin.py
Khorkhe: after meeting, it was decided that the risk caused by this ambiguity is too low to implement the complex solution involving traversing the MRO.

src/main/python/smv/stacktrace_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+
+FUNC_LIST = []
+
+class WithStackTrace(object):
+    """abstract class that can be mixed in to allow functions to use the
+        'with_stacktrace' decorator.
+        This will enable proper error reporting in functions that can be overridden
+        by either end-users or developers
+    """
+    __metaclass__ = abc.ABCMeta
```
laneb: Why is this an ABC?

laneb:
Please add a note about the what may happen i the decorator is used on methods with the same name in different classes.

Khorkhe:
@laneb a doc file was added for this, under docs/dev/Python/mixins.md

laneb:
Kewl nevermind then.

src/main/python/smv/stacktrace_mixin.py
Khorkhe: added so one would not try to instantiate an object of this class. I can remove if you prefer

Issue #822, opened 2017-05-26
sbt pytest fails


ninjapapa:
```
======================================================================
ERROR: test_SmvCsvFileWithUserSchema (testSmvFramework.SmvFrameworkTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bzhang/Tresamigos/SMV/src/test/python/testSmvFramework.py", line 58, in test_SmvCsvFileWithUserSchema
    df = self.df(fqn)
  File "/Users/bzhang/Tresamigos/SMV/src/main/python/test_support/smvbasetest.py", line 67, in df
    return cls.smvApp.runModule("mod:" + fqn)
  File "/Users/bzhang/Tresamigos/SMV/src/main/python/smv/smvapp.py", line 153, in runModule
    jdf = self.j_smvPyClient.runModule(urn, forceRun, self.scalaOption(version))
  File "/Users/bzhang/spark-1.5.2/python/lib/py4j-0.8.2.1-src.zip/py4j/java_gateway.py", line 538, in __call__
    self.target_id, self.name)
  File "/Users/bzhang/spark-1.5.2/python/lib/pyspark.zip/pyspark/sql/utils.py", line 36, in deco
    return f(*a, **kw)
  File "/Users/bzhang/spark-1.5.2/python/lib/py4j-0.8.2.1-src.zip/py4j/protocol.py", line 300, in get_return_value
    format(target_id, '.', name), value)
Py4JJavaError: An error occurred while calling o1576.runModule.
: py4j.Py4JException: An exception was raised by the Python Proxy. Return Message: x
	at py4j.Protocol.getReturnValue(Protocol.java:417)
	at py4j.reflection.PythonProxyHandler.invoke(PythonProxyHandler.java:113)
	at com.sun.proxy.$Proxy33.instanceValHash(Unknown Source)
	at org.tresamigos.smv.SmvExtModulePython.instanceValHash(SmvDataSet.scala:946)
	at org.tresamigos.smv.SmvDataSet.datasetHash(SmvDataSet.scala:122)
	at org.tresamigos.smv.SmvDataSet.hashOfHash$lzycompute(SmvDataSet.scala:136)
	at org.tresamigos.smv.SmvDataSet.hashOfHash(SmvDataSet.scala:135)
	at org.tresamigos.smv.SmvDataSet.verHex(SmvDataSet.scala:210)
[...]

PR #833, opened 2017-05-30
always use test-specific temporary data dir for Python tests (#822)

laneb:
Fixing #822.

### DIFF
--- a/src/main/python/test_support/smvbasetest.py
+++ b/src/main/python/test_support/smvbasetest.py
@@ -22,6 +22,7 @@ from pyspark.sql import *
 import os, shutil
 
 class SmvBaseTest(unittest.TestCase):
+    """DataDir value is deprecated. Use tmpDataDir instead."""
     DataDir = "./target/data"
     PytestDir = "./target/pytest"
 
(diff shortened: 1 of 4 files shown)


laneb:
Need to fix something. Will reopen when ready for review.

src/main/python/test_support/smvbasetest.py
```diff
@@ -22,6 +22,7 @@
 import os, shutil
 
 class SmvBaseTest(unittest.TestCase):
+    """DataDir value is deprecated. Use tmpDataDir instead."""
```
AliTajeldin: The above comments makes the warning the doc string for SmvBaseTest.  Just use the `#` instead

src/main/python/test_support/smvbasetest.py
```diff
@@ -82,14 +83,22 @@ def tmpTestDir(cls):
         return cls.PytestDir + "/" + cls.__name__
 
     @classmethod
+    def tmpDataDir(cls):
```
AliTajeldin: Add comment about what this is for and why we need it (as opposed to good old datadir)

src/main/python/test_support/smvbasetest.py
laneb: Oh good catch. I meant for it to be the docstring for the attribute but that probably doesn't work if it's the first line of the class. I guess we're not documenting the `SmvBaseTest` API yet anyway.

src/main/python/test_support/smvbasetest.py
laneb: Fixed.

src/main/python/test_support/smvbasetest.py
laneb: Fixed.

Issue #171, opened 2015-08-26
allow user to specify spark home in run_app.sh and run_shell.sh


AliTajeldin:
By default the script use the spark-submit and spark-shell that is in the path.
Should allow user to override the spark home dir.  For example:

``` bash
$ run_app.sh --spark-home /tmp/spark-1.1 --output-dir output/1.1 -s stage1
```

laneb:
Is something like this still needed?

AliTajeldin:
The intent was for developers who might have multiple versions of spark around.  For example, I was just testing the 2.1 branch and I would have to run `smv-pyrun` from either the `master` or `2.1` branch (I cloned the 2.1 into another directory rather then `git checkout`).

I'm not sure if the explicit flag is still needed.  But we do need to confirm that if the user runs `smv-pyrun` from anther SMV directory other than the one in the path, it runs everything from that new directory (jar files, executables, etc).

laneb:
Having the `--spark-home` option will be helpful when developing features across multiple Spark versions. Will add. Can confirm the second point described.

laneb:
Done.

PR #841, opened 2017-05-31
add option for user-specified spark home (#171)

laneb:
Addressing #171.

### DIFF
--- a/tools/_env.sh
+++ b/tools/_env.sh
@@ -19,8 +19,15 @@ function split_smv_spark_args()
             shift
             break
         fi
-        SMV_ARGS=("${SMV_ARGS[@]}" "$1")
-        shift
+
+        if [ "$1" == "--spark-home" ]; then
+          shift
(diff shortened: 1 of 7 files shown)


tools/_env.sh
```diff
@@ -75,14 +82,18 @@ function find_fat_jar()
 }
 
 function set_spark_home() {
-    if [ -z "$SPARK_HOME" ]; then
+    if [ -n "$SPARK_HOME_OPT" ]; then
+      SPARK_HOME="$SPARK_HOME_OPT"
```
AliTajeldin: This may need to be exported.  Line 93 below exports the setting of `SPARK_HOME` so I'm assuming it is needed by some child process.

laneb: Didn't notice the other export. This function was being used only by `smv-r`, but I will include the export in any case.

laneb: Fixed.

PR #852, opened 2017-06-01
#843: add .sh extension to smv-release, and exclude the admin directo…

Khorkhe:
…ry from the tar


AliTajeldin:
Please move script back to `smv-release`.  That is now the new naming convention to stay consistent with `smv-pyrun`, `smv-install`, etc.

Issue #392, opened 2016-10-06
Need a easy way to read persisted EDD (both Scala and Python sides)


ninjapapa:
When `genEdd` flag in the SmvApp is turned on, the Edd is calculated and persisted. However there is not easy method to read in the Edd result into shell and let user to browse through. Need to create a DS level method to support that, and also implement shell command to call that method.

laneb:
Is this still needed?

AliTajeldin:
Yes. I actually think this could be quite handy

PR #859, opened 2017-06-03
I392 read persisted edd

laneb:
Addressing #392. It's worth highlighting that the persisted EDD will lose some of its ordering due to partitioning - this was always the case for persisted EDD, but EDD in the shell will also lose its order. If necessary we could avoid this by collecting EDD in memory and exporting as a single CSV without losing the order. For now the EDD rows are sorted by the columns of the df the EDD is derived from.

### DIFF
--- a/src/main/python/smv/smvshell.py
+++ b/src/main/python/smv/smvshell.py
@@ -195,6 +195,14 @@ def discoverSchema(path, n=100000, ca=None):
     """
     SmvApp.getInstance()._jvm.SmvPythonHelper.discoverSchema(path, n, ca or SmvApp.getInstance().defaultCsvWithHeader())
 
+def edd(ds_name):
+    """Print edd report for the result of an SmvDataSet
+
+        Args:
+            ds_name (str): name of an SmvDataSet
+    """
+    _jvmShellCmd().edd(ds_name)
+
 def run_test(test_name):
     """Run a test with the given name without creating new Spark context
 
--- a/src/main/python/smv/smvshell.py
+++ b/src/main/python/smv/smvshell.py
@@ -201,7 +201,7 @@ def edd(ds_name):
         Args:
             ds_name (str): name of an SmvDataSet
     """
-    _jvmShellCmd().edd(ds_name)
+    print _jvmShellCmd()._edd(ds_name)
 
 def run_test(test_name):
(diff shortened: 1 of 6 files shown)


laneb:
The failing test won't be trivial to fix. Closing until it is fixed.

src/main/python/smv/smvshell.py
```diff
@@ -195,6 +195,14 @@ def discoverSchema(path, n=100000, ca=None):
     """
     SmvApp.getInstance()._jvm.SmvPythonHelper.discoverSchema(path, n, ca or SmvApp.getInstance().defaultCsvWithHeader())
 
+def edd(ds_name):
+    """Print edd report for the result of an SmvDataSet
+
+        Args:
+            ds_name (str): name of an SmvDataSet
+    """
+    _jvmShellCmd().edd(ds_name)
```
AliTajeldin: The convention is to have have the `jvmShellCmd`s return a string and do the printing here.  That way, the generated strings can be easily tested.  See `graphStage` above for an example.

ninjapapa: agreed.

laneb: Fixed.

src/main/scala/org/tresamigos/smv/shell/ShellCmd.scala
```diff
@@ -171,10 +171,11 @@ object ShellCmd {
...
     dsm.inferDS(name).head.exportToCsv(path)
 
-  def edd(name: String): Unit = {
-    val ds = dsm.inferDS(name).head
-    println(ds.getEdd)
-  }
+  def _edd(name: String): String =
+    dsm.inferDS(name).head.getEdd
+
+  def edd(name: String): Unit =
```
AliTajeldin: Is this still needed?

laneb: It can be omitted since we are not actively supporting the Scala shell. Would you prefer I remove it?

AliTajeldin: aaah, no it is ok.  I forgot that ShellCmd was used for the scala shell.

PR #863, opened 2017-06-06
I853 rm or fix sbt tests

laneb:
Tests removed:
  - `ModuleCrcConsistencyTest` (no longer useful)
  - `SmvExtModuleLinkTest` (targeted behavior no longer expected)
  - `SmvJdbcTest` case targeting append mode (doesn't work due to bug with JDBC and Derby)

Tests fixed:
  - `SmvFollowLinkTest` cases targeting `datasetHash`

This leaves several `ModuleCRCTest` cases and a `GraphUtilTest` case ignored. These will be fixed as part of other issues.


laneb:
Note: `SmvLockTest` failed with a spurious error we haven't seen before.

Issue #657, opened 2017-04-20
integrate gen-doc.sh with release_smv.sh


AliTajeldin:
Need to split `gen-doc.sh` into two parts.  One that is exclusively for generating the api doc and the other for publishing the docs.  The publishing part should be integrated with `release_smv.sh`.

jacobdr:
Looks like we should also standardize how we name our bash scripts...

PR #864, opened 2017-06-07
I657 gendoc in release

AliTajeldin:
#657 update `smv-install` to integrate generation and publishing of docs into release script.

PS: will need to delete the two old scripts once script has been tested on a real release.


laneb:
Looks good.

Khorkhe:
In order to release v1.6.2, I had to add py4j (pip install py4j) in the virtual environment. You may want to add that to the list of requirements. v1.5.2 does not require it though.

AliTajeldin:
@Khorkhe We should be using the py4j from the spark version (we add the following to python path in the release script `${spark_py_dir}/lib/py4j-0.8.2.1-src.zip`).  My guess is that the 1.6 version used a different py4j version which means we were not picking it up from spark lib dir.
I'll modify the code to use a glob pattern instead.

Khorkhe:
sounds good!

AliTajeldin:
Added glob pattern matching for py4j lib.

Issue #862, opened 2017-06-06
New SmvModule for modelling


laneb:
When modeling it would be helpful if a module could return a model as a result rather than a `DataFrame`. Need to create special module `SmvTraining` that returns a model as a result and`SmvModel` that receives a model as input.

AliTajeldin:
The classes should be `SmvModel` and `SmvModelExec`.  `SmvModel` is where the training takes place.

laneb:
Done.

PR #865, opened 2017-06-07
I862 smv result module

laneb:
This facilitates #862. `SmvModel` will implement `SmvResultModel`.


src/main/python/smv/smvdataset.py
```diff
@@ -601,12 +610,68 @@ def run(self, i):
...
+
+            Override this method to define the output of this module, given a map
+            'i' from inputSmvDataSet to resulting DataFrame. 'i' will have a
+            mapping for each SmvDataSet listed in requiresDS. E.g.
+
+            def requiresDS(self):
+                return [MyDependency]
+
+            def run(self, i):
+                return i[MyDependency].select("importantColumn")
```
ninjapapa: The example here is confusing. How about 
```
return train_model(i[MyDependency])
```

laneb: This whole docstring including the example is just c/p from `SmvDataSet.run`. I need to update the contents to make them more relevant.

laneb: Yes, forgot to update the example.

laneb: Fixed.

laneb:
See #869 for full PR.

PR #869, opened 2017-06-08
I862 smv model

laneb:
Completing #862. Most of the commits appearing right now are from i862_smv_result_module. I think this PR should shrink significantly once that branch is merged.


AliTajeldin:
Can you combine both into one PR.  Really hard and confusing to track which is which.  They are really all one issue anyway.

laneb:
I'll just close the other PR then.

docs/user/smv_dataset.md
```diff
@@ -174,3 +174,28 @@ class MyModule(SmvModule, SmvOutput):
     def run(self, i):
         ...
 ```
+
+## SmvModel and SmvModelExec
```
ninjapapa: I like the client code.
Should we make it a separate md and a separate session in the main user doc?

AliTajeldin: Yes indeed 👍  for the separate doc. 
May also want to expand on the introduction verbiage to indicate how to use the training / exec modules.

src/main/python/smv/smvdataset.py
```diff
@@ -600,13 +614,136 @@ def run(self, i):
                 (DataFrame): ouput of this SmvModule
         """
 
+    def _constructRunParams(self, known):
```
AliTajeldin: needs a quick method doc on how we are constructing the run params as it is not trivial.  Also, describe what is `known`.

src/main/python/smv/smvdataset.py
```diff
@@ -600,13 +614,136 @@ def run(self, i):
...
+        i = self._constructRunParams(known)
         return self.run(i)
 
+class SmvResultModule(SmvModule):
+    """An SmvModule whose result is not a DataFrame
+
+        The result must be picklable - see https://docs.python.org/2/library/pickle.html#what-can-be-pickled-and-unpickled.
+    """
+    @classmethod
+    def df2result(self, df):
```
AliTajeldin: also need to add a comment as to why we are using hexlify.  might be a code comment rather than the method doc.
Ooops, documented below.  Just reference the `result2df` method for explanation about hexify.

laneb: Done.

src/main/python/smv/smvdataset.py
laneb: Done.

docs/user/smv_dataset.md
laneb: Done.

AliTajeldin:
@laneb the branch has conflicts with master now.

laneb:
Yeah there are some complications with Python3. The `cPickle` module was removed and it looks like `binascii.hexlify` changed to.

laneb:
Merge conflicts and Python 3 issues are fixed.

Issue #866, opened 2017-06-07
Attach @with_stacktrace to all ISmvModule methods


laneb:
Some are still getting buried right now.

laneb:
@Khorkhe I've already made a PR for this. Should've assigned myself, sorry.

Khorkhe:
Oh whoops, will revert my local changes then

laneb:
Done.

PR #870, opened 2017-06-08
I866 fix stacktraces

laneb:
Addressing #866. 

Also, after reflecting on this design a bit, I think there is significant risk of printing redundant stack traces if code changes significantly in the future. Particularly, the current design doesn't account for  one @with_stacktrace method calling another, or if a @with_stacktrace method has an error when called by the Python driver (rather than as a callback). Seems low priority to fix these problems though.


AliTajeldin:
@laneb Dont forget to add the issue number to the commits.

PR #874, opened 2017-06-09
organize Python resources (#792)

laneb:
Addressing #792.


Khorkhe:
nice, this seems to standardize on a lot of things for how tests will be written.

Issue #137, opened 2015-07-28
add command line option to discover dead modules.


AliTajeldin:
Once stages are fully implemented (issue #126), we will be able to determine if any `SmvModule` in any of the packages of a stage is actually being used to affect one of the output modules of the same stage.  If a module is not in any of the dependency graph of any of the output modules, then it is a dead module.

Add a command line flag (e.g. "--dead") that will dump the list of all such dead modules.

laneb:
Still needed?

AliTajeldin:
yes

laneb:
Done.

PR #877, opened 2017-06-13
add option to print dead modules from commandline (#137)

laneb:
Addressing #137.


AliTajeldin:
Need to add the flag to the user doc.

laneb:
It felt like I was forgetting something...

laneb:
Fixed.

Issue #857, opened 2017-06-02
add official support for python 3.x


AliTajeldin:
The internals should already work and this might just be relaxing the check in `smv-install` (hopefully 🤞 )

laneb:
Should the Python version for Travis CI be updated?

jacobdr:
I think we should use a tool like tox and test both versions.

Jacob

AliTajeldin:
For travis, I think it is as easy as adding the following to the yml file:
```
python:
  - "2.7"
  - "3.2"
```

AliTajeldin:
Also, Travis has support for user defined matrix (you can mix and match multiple languages).  I think this is better than just adding python modules so that we can just repeat the python tests with each version and not have to re-run the scala tests.
See https://stackoverflow.com/a/44054333/7753976 for an example.

laneb:
Based on [this issue](https://github.com/travis-ci/travis-ci/issues/4090) it doesn't appear that travis 's default build image supports combining multiple language requirements - you can test against multiple versions of a single language, but not multiple languages as needed for the Python tests. The suggestion made in that issue is to use the Trusty image which is more flexible.

laneb:
Unfortunately, I get the following error when trying to run the Python unit tests with Python 3.5.3:
```
Spark Command: /Library/Java/JavaVirtualMachines/jdk1.8.0_121.jdk/Contents/Home/bin/java -cp target/scala-2.10/smv-1.5-SNAPSHOT-jar-with-dependencies.jar:/Users/laneb/Library/spark/conf/:/Users/laneb/Library/spark/lib/spark-assembly-1.5.2-hadoop2.7.2.jar:/Users/laneb/Library/spark/lib/datanucleus-api-jdo-3.2.6.jar:/Users/laneb/Library/spark/lib/datanucleus-core-3.2.10.jar:/Users/laneb/Library/spark/lib/datanucleus-rdbms-3.2.9.jar -Xms1g -Xmx1g org.apache.spark.deploy.SparkSubmit --conf spark.driver.extraClassPath=target/scala-2.10/smv-1.5-SNAPSHOT-jar-with-dependencies.jar --jars target/scala-2.10/smv-1.5-SNAPSHOT-jar-with-dependencies.jar, /Users/laneb/twine/SMV/tools/../src/main/python/scripts/runtests.py
========================================
Traceback (most recent call last):
[...]

PR #879, opened 2017-06-13
Support for Python 3

laneb:
Addressing #857.


jacobdr:
Looks like this is starting to shape up. Strong work.

laneb:
There were some issues with inconsistent whitespacing on master, so the PR test was failing. Only seemed to be a problem for Python3. Fixed now.

Issue #868, opened 2017-06-07
schema csv attributes update is not reflected


XiaosenChen:
Module A (a SmvCsvFile) points to a csv file whose schema CSV attributes has been updated. Running Module A again after the update does not reflect the change.
currently, user can manually purge this data in the ouput dir to solve this problem.

Khorkhe:
Hi @XiaosenChen 
I've attempted to reproduce this issue, as follows:

 - SmvCsvFile points to some csv file. The data dir also contains a .schema file for this csv
 - Run the first time, it persists: `16:48:17 PERSISTING: file:///Users/khorkhe/Desktop/temp/enter/data/output/stage1.employment.EmploymentByState_f68542ef.csv`
 - Edit the .schema file (say, change a coltype from String to Integer)
 - Run the second time again. It persisted under a different hash: `16:49:44 PERSISTING: file:///Users/khorkhe/Desktop/temp/enter/data/output/stage1.employment.EmploymentByState_b4d9346e.csv`

Am i missing something?

XiaosenChen:
try changing the csv attributes such as quote_char

Khorkhe:
done

PR #887, opened 2017-06-14
#868: add schema modification time to hash to handle schema attributes

Khorkhe:


src/main/scala/org/tresamigos/smv/SmvDataSet.scala
```diff
@@ -653,15 +653,18 @@ abstract class SmvFile extends SmvInputDataSet with SmvDSWithParser {
    *  - input csv file path
    *  - input csv file modified time
    *  - input schema contents
```
Khorkhe: is it necessary to hash the content of the schema file now that the schema file date is added to the hash? or would you prefer removing this?

AliTajeldin: was it hashing the contents of the schema file before?

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
```diff
@@ -653,15 +653,18 @@ abstract class SmvFile extends SmvInputDataSet with SmvDSWithParser {
...
+   *  - input schema file modified time (to handle attribute change)
    */
   override def instanceValHash() = {
     val fileName = fullPath
+    val schemaFileName = fileName.substring(0, fileName.lastIndexOf('.')) + ".schema"
```
AliTajeldin: That is already computed in `finalSchemaPath`.

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
laneb: Yes, we started hashing the schema contents when we added the `userSchema` method for specifying schema in the module code. Current implementation is a CRC of the `toString` of the schema. `toString` must not include the `CsvAttributes`.

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
Khorkhe: yes it was fixed shortly after the initial commit.

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
AliTajeldin: Then the real fix for this is to fix the `toString` method of the schema.  It should include the CsvAttributes.  Then, we will not need the file timestamp and the hash should work for both file schema and code schema.

src/main/scala/org/tresamigos/smv/SmvSchema.scala
```diff
@@ -347,7 +347,17 @@ class SmvSchema(val entries: Seq[SchemaEntry], val attributes: Map[String, Strin
 
   private[smv] def toValue(ordinal: Int, sVal: String) = entries(ordinal).typeFormat.strToVal(sVal)
 
-  override def toString = "Schema: " + entries.mkString("; ")
+  override def toString = {
+    val attrStr = if (attributes.size > 0) {
```
laneb: I think it would be better for this logic to be encapsulated in a `toString` method of `CsvAttributes`.

laneb:
Please include a unit test that changing the `CsvAttributes` changes the hash.

AliTajeldin:
Rather than modify `toString`, we should probably just use `toStringsWithMeta` as that already has the attributes and meta data in addition to the individual entries.
No need to modify `CsvAttributes`.  That task is a separate ticket (cleaning up the attributes in SmvSchema would require creating an `SmvAttributes` class that is the base of `CsvAttributes` and have that be the member of `SmvSchema`)
Agree on the need for unit test.

AliTajeldin:
PS: still missing the test.

Khorkhe:
Yes, will do later tonight

AliTajeldin:
If the work is not complete, close the PR and reopen once all check-ins are complete.  When it is open, it is on my list to review.

Khorkhe:
ready for review

Issue #884, opened 2017-06-13
Command line --publish-local behaves differently when the module is persisted


ninjapapa:
--publish-local will "publish" the csv without header when the resolved module already got persisted. However the "published" csv will have a header when the module was not persisted.

laneb:
Reviewing the code for` --publish-local`:

- for a non-ephemeral module that has been persisted, we just do a copy-merge

- for a non-ephemeral module that has not been persisted, persist it and then do a copy-merge, omitting the header

- for ephemeral modules we use `SmvDFHelper.smvExportCsv`, which writes the header 

Is the desired behavior to always write the schema in the header?

ninjapapa:
The `smvExportCsv(module_name, path)` method in shell also has this inconsistency. At least the `smvExprortCsv` function should behave the same as `df.smvExportCsv(file)` method, which is to write **with** header.

If we consider `--publish-local` is the same as `smvExportCsv`, it should also write with header. The only concern there is that `--publish-local` is not `--publish` on local.

AliTajeldin:
yes. `publish-local` should always have the headers.  Not sure what you mean by "The only concern there is that --publish-local is not --publish on local"

laneb:
@ninjapapa mentioned to me that there may be a use case for publishing a headerless csv and schema file locally. It seems that it may be appropriate to distinguish an `export-csv` whose result is a CSV with header, and a `publish-local` whose result is a headerless CSV and schema file.

AliTajeldin:
I think by default, if we are creating a single file, it should have a header.  The only reason we remove the header from the intermediate files is because there is was no way to guarantee that the header line would be the first line when generating a multi-partition file.  It should be possible to guarantee header when creating a single file.
If we need header-less output files, we should add an option in the future.

laneb:
For now, I will change `SmvDataSet.smvExportCsv` to include the header, which will fix both exporting from the shell and `publish-local`. The appropriate name for the option currently called `publish-local` is a separate conversation.

laneb:
@ninjapapa have you tried the using the toLocalIterator [method](https://stackoverflow.com/questions/21698443/spark-best-practice-for-retrieving-big-data-from-rdd-to-local-machine) mentioned in this issue? Description in Spark API doc is 

> Return an iterator that contains all of the elements in this RDD.
> 
> The iterator will consume as much memory as the largest partition in this RDD.

[...]

PR #891, opened 2017-06-15
I884 publish local with header

laneb:
Addressing #884.

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SmvReportIO.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvReportIO.scala
@@ -33,7 +33,7 @@ private[smv] object SmvReportIO {
     val pw = new PrintWriter(outFile)
     // According to Spark API doc, RDD.toLocalIterator will consume no more
     // memory than what is required for the largest partition
-    report.toLocalIterator foreach (r => pw.write(r.toString + "\n"))
+    report.toLocalIterator foreach (s => pw.write(s + "\n"))
     pw.close
   }
 
--- a/src/main/scala/org/tresamigos/smv/SmvReportIO.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvReportIO.scala
@@ -17,24 +17,34 @@ package org.tresamigos.smv
 import org.apache.spark.rdd.RDD
 import java.io.{PrintWriter, File}
 
+private[smv] class ReportWriter(path: String) {
+  val outFile = new File(path)
+  val parent = outFile.getParentFile
+  if (parent != null && !parent.exists)
+    throw new SmvRuntimeException(s"Cannot write to ${path}: directory ${parent} does not exist")
+  val pw      = new PrintWriter(outFile)
+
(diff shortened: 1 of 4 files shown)


src/main/scala/org/tresamigos/smv/SmvReportIO.scala
```diff
@@ -30,16 +30,11 @@ private[smv] object SmvReportIO {
...
-
-    // Save report RDD to a temparory file on HDFS
-    val tmpHdfsFile = "/tmp/smv_tmp_" + filename
-    SmvHDFS.deleteFile(tmpHdfsFile)
-    report.saveAsTextFile(tmpHdfsFile)
-
-    // copy merge the HDFS output to a local output
-    SmvHDFS.copyMerge(tmpHdfsFile, path)
-    SmvHDFS.deleteFile(tmpHdfsFile)
+    val pw = new PrintWriter(outFile)
```
AliTajeldin: Where is the header being written?

src/main/scala/org/tresamigos/smv/SmvReportIO.scala
```diff
@@ -30,16 +30,11 @@ private[smv] object SmvReportIO {
...
+    // According to Spark API doc, RDD.toLocalIterator will consume no more
+    // memory than what is required for the largest partition
+    report.toLocalIterator foreach (r => pw.write(r.toString + "\n"))
```
AliTajeldin: I'm assuming the type of `r` is `Row`.  Is `r.toString` taking care of quoting special characters?

src/main/scala/org/tresamigos/smv/SmvReportIO.scala
laneb: The header is the first entry of the RDD.

src/main/scala/org/tresamigos/smv/SmvReportIO.scala
laneb: Actually, the type of `r` is `String` and does not need the additional `toString`. Will fix.

laneb:
Just noticed a behavior change on this branch. Hadoop was automatically creating directories for us if we tried to copy-merge into a directory that didn't exist. With the current implementation, there will be an error if the directory doesn't exist. Need to fix that.

laneb:
Actually, per @AliTajeldin the behavior should not be to create the directory automatically. We will verify that the directory exists and throw an error otherwise.

src/main/scala/org/tresamigos/smv/SmvReportIO.scala
```diff
@@ -17,24 +17,34 @@ package org.tresamigos.smv
 import org.apache.spark.rdd.RDD
 import java.io.{PrintWriter, File}
 
+private[smv] class ReportWriter(path: String) {
```
AliTajeldin: guess what's missing from this class definition :smile:

laneb: -_-

PR #897, opened 2017-06-17
I896 Add help option to smv-init + various other improvements

### LINKED ISSUE #896: Add help `-h` option to smv-init
When I ran `smv-init -h`, I get a bazillion errors. While we have a help message, because `-h` is 1 argument, the [bad usage message](is not invoked)

jacobdr:
Add help option (-h and --help) to smv-init. Fixes #896 

Other major feature is adding Scala files to the enterprise mode. I think this was the intended behavior -- but for some reason enterprise mode previously meant that only a `smv-user-conf.props` file was additionally included.

### DIFF
--- a/tools/smv-init
+++ b/tools/smv-init
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+QUITE_MODE=0
+if [ "$1" = "-q" ]; then
+    QUITE_MODE=1
+    shift
+fi
+
 TEMPLATE_DIR="$(cd "`dirname "$0"`"; pwd)/templates"
 TEMPLATE_DATA_DIR="$TEMPLATE_DIR/data"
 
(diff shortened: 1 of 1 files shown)


tools/smv-init
```diff
@@ -121,7 +121,7 @@ function copy_conf_files()
     # copy appropriate files based on project type
     PYTHON_FILES="README.md notebooks/Untitled.ipynb .gitignore.template log4j.properties conf/smv-app-conf.props conf/smv-user-conf.props"
     SCALA_FILES="$PYTHON_FILES build.sbt pom.xml project/assembly.sbt project/build.properties"
-    ENTERPRISE_FILES="$PYTHON_FILES conf/smv-user-conf.props.template"
+    ENTERPRISE_FILES="$PYTHON_FILES $SCALA_FILES conf/smv-user-conf.props.template"
```
laneb: The Scala files were intentionally removed a while back and new users are discouraged from using the Scala API. Can you explain the reasoning for including them here?

jacobdr:
In the simple setting -- the default -- that makes total sense to me. But what, then, was the point of the enterprise option? Before this change it just added a user-conf file. That seems fairly pointless to me.

I actually have a project now that involves interest in the Scala side, so having this helper returned is useful.

Jacob

AliTajeldin:
@jacobdr : This PR should be limited to adding support for `-h` help message.  If we need to add scala back, we need a different ticket to discuss.  We took out the scala files on purpose.  If we do need them, then perhaps we need a special flag for that (e.g. `-scala` flag to `smv-init`)
BTW, the enterprise was to create an app with multiple stages and have the input definition separate from the module definition for scalability.

AliTajeldin:
PS: make sure your checkins have the issue number in them.  Generally prefer the form "#xxx: description".

AliTajeldin:
@jacobdr we should either close this PR forever and create a new PR for only the help function or update the code to only have the help function and not the scala changes.

AliTajeldin:
@jacobdr bump

jacobdr:
I still need to fix this I messed up the rebase (naturally).

AliTajeldin:
@jacobdr has this been fixed?  Is this ready for merge?

jacobdr:
I think it needs a rebase or at least a little more testing... please leave it open for now

jacobdr:
Way too out of date....

AliTajeldin:
@jacobdr might be easier to just apply the deltas from the above changes by hand rather than trying an automated merge.

PR #898, opened 2017-06-17
Add travis caching to speed up build times

jacobdr:
Stash the ivy repo between builds to speed up build times. 

I am not 100% sure this is going to work as some of the documentation seems to indicate that we need to be on the the Travis "container-based infrastructure", and I am not 100% sure that happens since we run Travis with the `sudo: required` option. 

Docs:
* http://www.scala-sbt.org/0.13/docs/Travis-CI-with-sbt.html#Caching
* https://docs.travis-ci.com/user/caching/
* Example from Spark [source](https://github.com/apache/spark/blob/master/.travis.yml)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,8 @@ metastore_db/
 
 .sync-config.cson
 *.pyc
+
+.bashrc
+.ivy2
+.jupyter/
+.sbt/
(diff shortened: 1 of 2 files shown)


laneb:
this looks slick

AliTajeldin:
@jacobdr Actually, I don't want to add complexity to our travis setup when it is not needed.  A few minutes build time is not hurting us.  Let's keep the branch incase we need to add caching in the future but let's close this PR.

jacobdr:
@AliTajeldin With all the recent build activity, any interest in reviving this?

AliTajeldin:
No.

PR #906, opened 2017-06-20
i901 Added deleting 'org' from sys.modules

ninjapapa:


ninjapapa:
Yes. The issue is still open. Need to review our dynamic importing and python's new import mechanism to make sure things will still work. 

For question 1, according to my understanding of Python's document, yes, if you imported `smv.shell`, you also imported `smv`. However to use things in `smv`, you need to explicitly put `smv.` as prefix, so user should not accidentally using things in `smv.

For question 2, current fix is for the loading Python project form `org.x.y` namespace. Since `import org.x.y` will really do the file search from `sys.path` when `org` is not in `sys.modules`.

Issue #901, opened 2017-06-19
Integration test failed on Py3.5 + 2.1 


ninjapapa:
Error message:
```
Starting Py4j callback server on port 25334
Traceback (most recent call last):
  File "/home/travis/build/TresAmigosSD/SMV/src/main/python/smv/stacktrace_mixin.py", line 20, in catch_errs
    res = func_needing_stacktrace(*pargs, **kwargs)
  File "/home/travis/build/TresAmigosSD/SMV/src/main/python/smv/datasetrepo.py", line 57, in loadDataSet
    ds = for_name(fqn)(self.smvApp)
  File "/home/travis/build/TresAmigosSD/SMV/src/main/python/smv/utils.py", line 29, in for_name
    mod = __import__(modname)
ImportError: No module named 'org.tresamigos'
Traceback (most recent call last):
  File "/home/travis/build/TresAmigosSD/SMV/tools/../src/main/python/scripts/runapp.py", line 19, in <module>
    app.run()
  File "/home/travis/build/TresAmigosSD/SMV/src/main/python/smv/smvapp.py", line 205, in run
    self.j_smvApp.run()
  File "/home/travis/build/TresAmigosSD/SMV/spark/python/lib/py4j-0.10.4-src.zip/py4j/java_gateway.py", line 1133, in __call__
  File "/home/travis/build/TresAmigosSD/SMV/spark/python/lib/pyspark.zip/pyspark/sql/utils.py", line 63, in deco
  File "/home/travis/build/TresAmigosSD/SMV/spark/python/lib/py4j-0.10.4-src.zip/py4j/protocol.py", line 319, in get_return_value
py4j.protocol.Py4JJavaError: An error occurred while calling o31.run.
: org.tresamigos.smv.SmvRuntimeException: SmvDataSet mod:org.tresamigos.smvtest.test5.modules.M1 not found
	at org.tresamigos.smv.DataSetResolver.org$tresamigos$smv$DataSetResolver$$findDataSetInRepo(DataSetResolver.scala:126)
	at org.tresamigos.smv.DataSetResolver$$anonfun$loadDataSet$1$$anonfun$apply$1.apply(DataSetResolver.scala:58)
	at org.tresamigos.smv.DataSetResolver$$anonfun$loadDataSet$1$$anonfun$apply$1.apply(DataSetResolver.scala:52)
	at scala.Option.getOrElse(Option.scala:121)
[...]

Issue #910, opened 2017-06-22
smv-install for 2.1.1.0 release installs spark 2.1.0


laneb:
Should install 2.1.1.

laneb:
Done.

Issue #909, opened 2017-06-22
Fix bug in smv-shell on 2.1 branch


laneb:
A `NameError` prevents importing the users shell init. Looks like a mistake was made when resolving a merge conflict.

laneb:
Actually, this has already been fixed on the branch. It just wasn't fixed before the last 2.1 release.

PR #911, opened 2017-06-22
remove 'org' from name of integration test modules

laneb:
Addressing #901. May only be needed for Spark 2.1 with Python 3.5 but the tests should be kept in sync across branches where possible. Can be merged into 2.1 after PR is merged.


AliTajeldin:
@laneb please merge master into 2.1 branch in prep for release 2.1.1.1 (once #910 and #909 are also completed)

Issue #921, opened 2017-07-12
Add a method to get SmvDataSet.needsToRun and pass it to the smvserver


tianruiwang:
Currently `needsToRun` is a private lazy val defined in the abstract class `SmvDataSet`. Need to pass it to the `smvserver.py` (perhaps through `SmvApp.getInstance().get_graph_json()`) s.t. pluto can get the run state information of each module.

PR #922, opened 2017-07-13
#921 pass needsToRun to the smv server

tianruiwang:
#921 This will pass the boolean value `needsToRun` of each module to smv-server through `get_graph_json()`.


AliTajeldin:
@tianruiwang worlds shortest PR 😄 😄 😄

Issue #928, opened 2017-07-20
New dsType for SmvModel and SmvModelExec


chengfei0913:
Currently, the dsType of SmvModel and SmvModelExec is "Module" which is extended from SmvModule.

We need to give them new dsTypes to distinguish them from smvModule

chengfei0913:
I will open a PR for this once I have the permission to push code

AliTajeldin:
Done

PR #930, opened 2017-07-21
#928 add ds type for smvModel and smvModelExec

chengfei0913:
#928 
Overrided dsType() function for smvModel and smvModelExec

### DIFF
--- a/src/main/scala/org/tresamigos/smv/graph/SmvStageGraph.scala
+++ b/src/main/scala/org/tresamigos/smv/graph/SmvStageGraph.scala
@@ -219,7 +219,10 @@ private[smv] class SmvGraphUtil(app: SmvApp, pstages: Seq[String] = Nil) {
 
     def toName(ds: SmvDataSet) = s"""\"${ds.fqn}\""""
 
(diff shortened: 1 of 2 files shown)


src/main/scala/org/tresamigos/smv/graph/SmvStageGraph.scala
```diff
@@ -219,18 +219,23 @@ private[smv] class SmvGraphUtil(app: SmvApp, pstages: Seq[String] = Nil) {
 
     def toName(ds: SmvDataSet) = s"""\"${ds.fqn}\""""
 
-    def toNodeStr(nodeType: String)(m: SmvDataSet) =
+    def toNodeStr(m: SmvDataSet) = {
+      val dsType = m.dsType
+      val nodeType = if (dsType == "Input") "file" else (
```
chengfei0913: I am curious why we use "file" instead of "input" as the node type

AliTajeldin: probably due to historic reasons as the only input used to be files.

PR #942, opened 2017-08-02
Support for pushing code with SmvGit

laneb:
Currently supports authorization through OAuth token, but we can adapt to support other authorization/authentication protocols as needed.

### DIFF
--- a/src/main/python/scripts/smvserver.py
+++ b/src/main/python/scripts/smvserver.py
@@ -213,6 +213,11 @@ def addFile(author, authorEmail, filePath, commitMessage, workDir="."):
     j_smvPyClient = SmvApp.getInstance().j_smvPyClient
     j_smvPyClient.addFile(author, authorEmail, filePath, commitMessage, workDir)
 
+def pushToRemote(workDir=".", authToken=None):
+    app = SmvApp.getInstance()
+    authTokenOpt = app.scalaOption(authToken)
+    app.j_smvPyClient.pushToRemote(workDir, authTokenOpt)
+
 # Wrapper so that other python scripts can import and then call
 # smvserver.Main()
 class Main(object):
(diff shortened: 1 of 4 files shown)


src/main/python/scripts/smvserver.py
```diff
@@ -213,6 +213,11 @@ def addFile(author, authorEmail, filePath, commitMessage, workDir="."):
     j_smvPyClient = SmvApp.getInstance().j_smvPyClient
     j_smvPyClient.addFile(author, authorEmail, filePath, commitMessage, workDir)
 
+def pushToRemote(workDir=".", authToken=None):
```
AliTajeldin: function comment

src/main/scala/org/tresamigos/smv/git/SmvGit.scala
```diff
@@ -41,15 +43,34 @@ case class SmvGit(workDir: String = ".") {
         .setAuthor(author, authorEmail)
         .setMessage(commitMessage)
         .call()
+    }
+
+  def addRemote(remoteName: String, remoteUrl: URL): Unit = {
```
AliTajeldin: yup, function comment.  when should this be called, how?  does the order matter?

src/main/scala/org/tresamigos/smv/git/SmvGit.scala
```diff
@@ -41,15 +43,34 @@ case class SmvGit(workDir: String = ".") {
...
+    val localConfig = withRepo(workDir) { _.getConfig }
+    val remoteConfig = new RemoteConfig(localConfig, remoteName)
+    val uri = new URIish(remoteUrl)
+    remoteConfig.addURI(uri)
+    remoteConfig.update(localConfig)
+    localConfig.save()
   }
 
+  def pushToRemote(remoteName: String = "origin", optAuthToken: Option[String] = None): Unit =
```
AliTajeldin: same

AliTajeldin:
might want to just close this PR for now while investigating pure python solution.  If we go back to this, then we can reopen and apply code changes requested here.

laneb:
Because external requirements for git support have changed, this PR is no longer necessary. We will probably remove SmvGit altogether. Closing.

Issue #953, opened 2017-08-09
Add an interface in SmvApp to get the SmvApp's config


guangningyu:
I'm working on a client application and need to get the SMV project's configuration (`dataDir`, to be precise).

## Current Design

Currently, there is an interface for each variable in the config in SmvApp class as below:

```python
class SmvApp(object):
    ...
    def appName(self):
        return self.j_smvApp.smvConfig().appName()
    ...
```

## Option 1

We expose the entire smvConfig() object to the users:

```python
class SmvApp(object):
    ...
    def smvConfig(self):
        return self.j_smvApp.smvConfig()
    ...
```

In this way, the users can/should get the specific variable in the config by themselves:
```python
appName = SmvApp.getInstance().smvConfig().appName()
dataDir = SmvApp.getInstance().smvConfig().dataDir()
```

## Option 2

We only expose the selected config variables in SmvApp class and add an interface for each of them.

```python
class SmvApp(object):
    ...
    def appName(self):
        return self.j_smvApp.smvConfig().appName()

    def dataDir(self):
        return self.j_smvApp.smvConfig().dataDir()
    ...
```

In this way, the users should call the corresponding interface:
```python
appName = SmvApp.getInstance().appName()
dataDir = SmvApp.getInstance().dataDir()
```

Which way is better?

jacobdr:
+1 for option 1

Jacob

AliTajeldin:
I'm curious where we need the data dir.  Is it from within the run method? Within test code?  I'm a bit worried about exposing the entire config as we would then be bound by that interface.  I just want to determine the use case first before discussing the how.

guangningyu:
The use case is to add a custom API in `smvserver` to get the file names/paths of all the input files. In this way, the client will be able to display a list of existing input files when a user is trying to create an input module.

The solution I'm considering is:
1. get the input directory, i.e. `dataDir`/input
2. return the list of files in the given directory

Is there a better way to do this?

laneb:
You might also consider getting a list of all the `SmvCsvFiles` in the project and mapping them to their paths. An advantage (or disadvantage) of this approach is that you would only see the files that correspond to existing `SmvCsvFiles`.

AliTajeldin:
In this particular use case, we need to see all files and not just the ones with existing `SmvCsvFile`

guangningyu:
@laneb Thanks, while this is not the case for me. As @AliTajeldin said, I need to see all the CSV files in the input directory.

[...]

PR #957, opened 2017-08-13
I953 return csv files in input dir

guangningyu:
#953


guangningyu:
@jacobdr 
1. Good point. Does `getFileNamesByType` make more sense?
2. I need to use the method `SmvHDFS.dirList` and I think `SmvApp` is the only interface to call Scala methods. That's why I add a wrapper here. Is there a better way to do this?

AliTajeldin:
@guangningyu lets go ahead with `getFileNamesByType`

PR #975, opened 2017-09-05
I953 return csv files in input dir

guangningyu:
#953


AliTajeldin:
This looks like it was already merged onto the 1.5 branch.  We shouldn't merge the same code onto both release branches.  We should just merge master into 2.1 (which we would need to do anyway before making 2.1 the master).
Please close unless there was something else that was needed here.

guangningyu:
@AliTajeldin Got it. Will close this PR.

Issue #961, opened 2017-08-22
Need a method in shell to return the "model" from an SmvModel


ninjapapa:
One may want to define an `SmvModel` in code, and play with it in shell. In that case, if we used `df` to resolved the model, a DF with pickled value will be returned, which is hard to use. Need a method to return the python object itself.

PR #976, opened 2017-09-14
add smv-pyshell function for getting SmvModel result (#961)

laneb:
#961

### DIFF
--- a/src/main/python/smv/smvshell.py
+++ b/src/main/python/smv/smvshell.py
@@ -37,6 +37,21 @@ def df(name, forceRun = False, version = None):
     """
     return SmvApp.getInstance().runModuleByName(name, forceRun, version)
 
+def getModel(name, forceRun = False, version = None):
+    """Get the result of running the named SmvModel module
+
+        Args:
+            name (str): The name of a module. Does not have to be the FQN.
+            forceRun (bool): True if the module should be forced to run even if it has persisted output. False otherwise.
+            version (str): The name of the published version to load from
+
+        Returns:
+            (object): The result of running the named module
+    """
+    app = SmvApp.getInstance()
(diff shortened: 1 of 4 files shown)


src/main/python/smv/smvshell.py
```diff
@@ -37,6 +37,21 @@ def df(name, forceRun = False, version = None):
     """
     return SmvApp.getInstance().runModuleByName(name, forceRun, version)
 
+def getModel(name, forceRun = False, version = None):
```
ninjapapa: Can we make `df` itself return model when call on an `SmvModel`? 
Regardless how we implement `SmvModel`, from an user's angle, he expect to get whatever he returned from the run method from `df`.

laneb: I didn't change `df` because it stands for `DataFrame` - it seemed a little confusing it for it to return other results. But it doesn't make much of a difference, because `getModel` should actually be able to regular `SmvModule` results right now. I can rename `getModel` to `df` and remove the old `df`. It will just mean that we can't use `df` to examine the `DataFrame` containing the pickled model.

laneb:
The build failure is just a spurious lost executor on travis.

ninjapapa:
No urgency to decide. Let's wait for Ali. Please add Ali as a reviewer.

src/main/python/smv/smvshell.py
AliTajeldin: I'm actually torn between the two options.  Neither seems ideal.  The only option I don't like is renaming `getModel` to `df` as it does work for normal modules but only because of accident.  Very confusing for future readers.
Let's stick with what we have for now.

laneb:
@AliTajeldin Done.

Issue #979, opened 2017-09-15
Create SmvSqlCsvFile which supports SQL commands


cloudysail:
Like `SmvSqlModule`, we want to support sql for the input `SmvCsvFiles`.

The user code looks like this:
```
class MyInput(SmvSqlCsvFile):
    def path(self):
        return "checkin.csv"
    def tableName(self):
        return "t1"
    def tableQuery(self):
        return "select a, b from t1"
```

laneb:
I edited your comment to remove reference to an issue in a private repo.

AliTajeldin:
@cloudysail Let's hardcode the tablename as there is only one table for the input.

PR #980, opened 2017-09-19
#979: create SmvSqlCsvFile

cloudysail:
#979

### DIFF
--- a/src/main/python/smv/smvdataset.py
+++ b/src/main/python/smv/smvdataset.py
@@ -388,36 +388,29 @@ class SmvSqlCsvFile(SmvCsvFile):
     """Input from a file in CSV format and using a SQL query to access it
     """
 
-    @abc.abstractmethod
-    def tableName(self):
-        """User-specified name for the table to extract input from the CSV file
+    # temporary table name
+    tableName = "df"
 
-            Override this to specify your own table name.
-
-            Returns:
-                (str): table name
-        """
-
-    def tableQuery(self):
+    def query(self):
         """Query used to extract data from the table which reads the CSV file
 
             Override this to specify your own query (optional). Default is
-            equivalent to 'select * from ' + tableName().
+            equivalent to 'select * from ' + tableName.
 
             Returns:
                 (str): query
         """
-        return "select * from "+ self.tableName()
+        return "select * from " + self.tableName
 
     def run(self, df):
         # temporarily register DataFrame of input CSV file as a table
-        df.registerTempTable(self.tableName())
+        df.registerTempTable(self.tableName)
 
         # execute the table query
-        res = self.smvApp.sqlContext.sql(self.tableQuery())
+        res = self.smvApp.sqlContext.sql(self.query())
 
         # drop the temporary table
-        self.smvApp.sqlContext.sql("drop table " + self.tableName())
+        self.smvApp.sqlContext.sql("drop table " + self.tableName)
 
         return res
 
--- a/src/main/python/smv/smvdataset.py
+++ b/src/main/python/smv/smvdataset.py
@@ -384,6 +384,43 @@ class SmvCsvFile(SmvFile):
                 (str): path
         """
 
+class SmvSqlCsvFile(SmvCsvFile):
+    """Input from a file in CSV format and using a SQL query to access it
+    """
+
+    @abc.abstractmethod
+    def tableName(self):
+        """User-specified name for the table to extract input from the CSV file
+
+            Override this to specify your own table name.
+
+            Returns:
+                (str): table name
+        """
+
+    def tableQuery(self):
+        """Query used to extract data from the table which reads the CSV file
+
+            Override this to specify your own query (optional). Default is
+            equivalent to 'select * from ' + tableName().
+
+            Returns:
+                (str): query
+        """
+        return "select * from "+ self.tableName()
+
+    def run(self, df):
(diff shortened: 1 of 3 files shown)


src/main/python/smv/smvdataset.py
```diff
@@ -384,6 +384,43 @@ def path(self):
...
+    def tableName(self):
+        """User-specified name for the table to extract input from the CSV file
+
+            Override this to specify your own table name.
+
+            Returns:
+                (str): table name
+        """
+
+    def tableQuery(self):
```
laneb: For consistency with `SmvSqlModule`, let's call it `query`.

src/main/python/smv/smvdataset.py
```diff
@@ -384,6 +384,43 @@ def path(self):
...
+
+            Override this to specify your own query (optional). Default is
+            equivalent to 'select * from ' + tableName().
+
+            Returns:
+                (str): query
+        """
+        return "select * from "+ self.tableName()
+
+    def run(self, df):
```
laneb: For DRY purposes, we should create a mixin for this logic which `SmvSqlModule` and `SmvSqlCsvFile` share. The only variance between the two implementations is the multiplicity of inputs.

src/main/python/smv/smvdataset.py
cloudysail: I did this because both`SmvHiveTable` and `SmvJdbcTable` use `tableQuery `.
Should we still use `query` here?

src/main/python/smv/smvdataset.py
cloudysail: Yes, I tried but later I found that it's hard to create a mixin.
We cannot use a for-loop over the run() here to replace the code in `SmvSqlModule `, because we have to use one for-loop to register tables first, and then execute SQL, and last use another for-loop to drop all temp tables.

src/main/python/smv/smvdataset.py
AliTajeldin: yes.  The word query here is overloaded but for the interface sake, `query` is for the spark code to run while `tableQuery` is for the code string to run on the remote system (hive or jdbc database).

src/main/python/smv/smvdataset.py
AliTajeldin: We could at a minimum create a mixin with two functions to register/cleanup temp tables. Not sure if it is worth the effort though.

src/main/python/smv/smvdataset.py
```diff
@@ -384,6 +384,43 @@ def path(self):
                 (str): path
         """
 
+class SmvSqlCsvFile(SmvCsvFile):
+    """Input from a file in CSV format and using a SQL query to access it
+    """
+
+    @abc.abstractmethod
+    def tableName(self):
```
AliTajeldin: Use a hardcoded name here.

src/main/python/smv/smvdataset.py
laneb: Rather than overloading `run`,  the mixin can give a method for executing the query, including registering/unregistering tables. It could take the dependency map as a parameter. Agreed it may not be worthwhile if this is the last dataset type we need a SQL variant of. We can keep this in our pocket until `SmvSqlMultiCsvFiles` comes around.

cloudysail:
Fix done. Please check.

By the way, one of the Travis build [919](https://travis-ci.org/TresAmigosSD/SMV/builds/277092052) failed. It looks like a previous build [905](https://travis-ci.org/TresAmigosSD/SMV/builds/277092052).
I saw comment from @laneb said "The build failure is just a spurious lost executor on travis." I wonder if it is same thing here which we can ignore?


```
 [info] - smvCountNull should count columns with null values *** FAILED ***
[info]   org.apache.spark.SparkException: Job aborted due to stage failure: Task 25 in stage 1494.0 failed 1 times, most recent failure: Lost task 25.0 in stage 1494.0 (TID 22217, localhost): ExecutorLostFailure (executor driver lost)
[info] Driver stacktrace:
[info]   at org.apache.spark.scheduler.DAGScheduler.org$apache$spark$scheduler$DAGScheduler$$failJobAndIndependentStages(DAGScheduler.scala:1283)
[info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$abortStage$1.apply(DAGScheduler.scala:1271)
[info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$abortStage$1.apply(DAGScheduler.scala:1270)
[info]   at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
[info]   at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
[info]   at org.apache.spark.scheduler.DAGScheduler.abortStage(DAGScheduler.scala:1270)
[info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$handleTaskSetFailed$1.apply(DAGScheduler.scala:697)
[info]   at org.apache.spark.scheduler.DAGScheduler$$anonfun$handleTaskSetFailed$1.apply(DAGScheduler.scala:697)
[info]   at scala.Option.foreach(Option.scala:236)
[info]   at org.apache.spark.scheduler.DAGScheduler.handleTaskSetFailed(DAGScheduler.scala:697)
```

laneb:
Yes, that just happens randomly on Travis sometimes. It's safe to ignore.

PR #1006, opened 2017-10-19
I1000 historical metadata

### LINKED ISSUE #1000: add historical validation ability to SmvDataset
Need to be able to perform validation based on historical data and not just the current run (as in DQM).  See design doc: https://github.com/TresAmigosSD/SMV/blob/master/docs/dev/Validator/validator.md

laneb:
Resolves #1000.

### DIFF
--- a/src/main/python/smv/smvdataset.py
+++ b/src/main/python/smv/smvdataset.py
@@ -279,6 +279,8 @@ class SmvDataSet(object):
         return {}
 
     def metadataJson(self, jdf):
+        """Get user's metadata and jsonify it for py4j transport
+        """
         df = DataFrame(jdf, self.smvApp.sqlContext)
         metadata = self.metadata(df)
         if not isinstance(metadata, dict):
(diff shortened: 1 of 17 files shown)


laneb:
Missed a detail we discussed today - need to rename the default metadata fields to avoid naming conflicts.

AliTajeldin:
FYI: test `test_invalid_metadata_validation_rejected_gracefully` is failing on Travis CI.

src/main/python/smv/smvdataset.py
```diff
@@ -263,6 +264,68 @@ def df2result(cls, df):
...
+
+            Override this method to define validation rules for metadata given
+            the current metadata and historical metadata.
+
+            Arguments:
+                current (dict): current metadata kv
+                history (list(dict)): list of historical metadata kv's
+
+            Returns:
+                (bool): True indicates success, False indicates failure
```
AliTajeldin: The type should be string or None.  A none-`None` value indicates failure.  `None` indicates sucess (or no return does the same).

PR #1008, opened 2017-10-21
propagate user SyntaxError that prevent modules from importing (#1004)

### LINKED ISSUE #1004: Python module syntax errors don't seem to get propagated
If, e.g., you forget a colon in your module definition, SMV doesn't propagate that error or stacktrace back to the Python driver.

laneb:
Fixes #1004. Note that I've reduced the generality of `utils.for_name`, but that generality was not useful to us. In fact, `for_name` is only used in a single place in the code.

### DIFF
--- a/src/main/scala/org/tresamigos/smv/DataSetRepo.scala
+++ b/src/main/scala/org/tresamigos/smv/DataSetRepo.scala
@@ -14,6 +14,8 @@
 
 package org.tresamigos.smv
 
+import scala.util.Try
+
 import classloaders.SmvClassLoader
 
 /**
@@ -21,7 +23,7 @@ import classloaders.SmvClassLoader
  * in a given language. A new repo is created for each new transaction.
  */
 abstract class DataSetRepo {
-  def loadDataSet(urn: ModURN): SmvDataSet
+  def loadDataSet(urn: ModURN): Option[SmvDataSet]
   def urnsForStage(stageName: String): Seq[URN]
 }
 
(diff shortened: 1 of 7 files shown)


src/main/scala/org/tresamigos/smv/DataSetRepo.scala
```diff
@@ -31,8 +33,10 @@ abstract class DataSetRepoFactory {
 
 class DataSetRepoScala(smvConfig: SmvConfig) extends DataSetRepo {
   val cl = SmvClassLoader(smvConfig, getClass.getClassLoader)
-  def loadDataSet(urn: ModURN): SmvDataSet =
-    (new SmvReflection(cl)).objectNameToInstance[SmvDataSet](urn.fqn)
+  def loadDataSet(urn: ModURN): Option[SmvDataSet] =
+    Try {
+      new SmvReflection(cl).objectNameToInstance[SmvDataSet](urn.fqn)
```
AliTajeldin: wouldn't this "eat up" any exceptions while loading the scala modules?  Or is the only possible exception on the scala side is if the module was not there?

laneb:
There is no risk (as far as I'm aware) of a runtime error in client code
causing an error during reflection. Syntax errors, undefined names, and
missing imports, which cause the bulk of errors during reflection on the
Python side, will prevent Scala code from compiling in the first place.
Correct me if I'm wrong, but I don't think any code will be executed as a
consequence of reflection in Scala (vs Python, where imports are executed
like a script on the first import). I suppose it is possible that there are
more esoteric errors that could occur during reflection, and I can look
into that if you want, but this approach for Scala code has been pretty
stable and precedes the DataSetManager.

AliTajeldin:
@laneb got it.  Even if there are some esoteric cases, we are not really using the scala interface that much anyway.

PR #1013, opened 2017-10-26
fix: accept alphanumeric chars in version and can compare reasonably

Kai-Chen:
A bit of an overkill :)

But it does provide a reasonably solid solution to compare versions.

All this might have been avoided if we have the GNU `sort` on a Mac, which has a `-V` switch to compare version numbers.
But the function here handles the `1.0rc1` variety corrrectly, while `sort -V` won't.


AliTajeldin:
@Kai-Chen Indeed, this is waaaaaaaay overkill.  We are just interested in getting the right spark version.  How about we use sed to extract the first 3 numeric "." separated fields and ignore everything else.

Kai-Chen:
Okay will do.

See what happens when I'm left with a bit of time on my own :)

Kai-Chen:
I left the tests in there so it's easier to update later.

The core of the logic is still the same: first split the fields by '.' and then zero-fill missing parts.
The individual part comparison is simpler now, because we know we're dealing with integers after sanitizing with sed.

AliTajeldin:
We are not going to maintain 60 lines of bash script code to compare versions.  Not going to happen, ever!   The cure is worse than the problem.  One of these days, this will fail on some weird client issue and I want something simple to debug and understand while the client is breathing down our neck.

We need to find a solution that is much simpler even if we have to make some compromise.  All I want to ensure is that the version matches the desired version.  We have a couple of options:
* How about we just match a prefix?  If the user specifies "1.5" then we match all "1.5.x" and if user specifies "1.5.2" than it has to match "1.5.2" exactly.
* If we want to support range, how about we just convert "1.5.2" into 010502 and assume versions and sub-versions are always less than two digits and do the compare that way.
* Another option is user provides a prefix plus minor number.  So the version match takes "1.5" and 2 and it will make sure the prefix is 1.5 and the minor number is at least 2.

Let's discuss the options above before we try to fix.

Kai-Chen:
I'm guessing it would take roughly the same number of lines to implement options 1 and 3.
In both cases, there's no getting around splitting by dot, and then compare each part as an integer, which is the bulk of the function.
The lengthiness is really not because the logic is convoluted, but because of the syntax of the language.

Option 2 should take fewer lines, and the assumption is reasonable.
But you'd need to grok some sed magic like the following

`sed 's/^[0-9]\./0&/; s/\.\([0-9]\)$/.0\1/; s/\.\([0-9]\)\./.0\1./g; s/\.\([0-9]\)\./.0\1./g'`

That should be okay after about 20 lines of comments explaining the regex :)

Another option would be to write or use a cmdline utility written in a more friendly language.
On Arch we have `vercmp` that does exactly that.
There's probably some python script also that should do this reasonably well.
Then we just add that to the build dependency and call it a day.

AliTajeldin:
How about:
* trim non version part of V:
```
$ echo "1.5.2howdy+12" | sed -e 's/\([0-9.]*\).*/\1/'
1.5.2
```

* split version into three parts:
```
$ IFS="." read -r -a a <<< "1.5.2"
$ echo ${a[@]}
1 5 2
```

* compare prefix:
```
$ [[ "1.5.2" == 1.5* ]]
$ echo $?
0
$ [[ "1.6.2" == 1.5* ]]
$ echo $?
1
```

Combining the three parts above should be ~ 6 lines of bash code to compare prefix + minor

Kai-Chen:
Yes, trimming and splitting is straight-forward.

The nearly-exact match would certainly be a lot simpler.
But how would you get the variable into the `[[ "1.5.2" == what_do_you_put_here* ]]` ?

Kai-Chen:
If the function is going to be used as in `ver_cmp(found="1.5.2", required="1.5")`, then there won't even be a need to split.
All we'd have to do is to trim and a * comparison.

Kai-Chen:
In that case, you wouldn't even need to trim.
A simple ``[[ "$installed" == ${required}* ]]` should be enough.

AliTajeldin:
I think the user will need to supply two values.  First, the major version match (e.g. "1.5") and the minimum minor version (e.g. "2").  This will match any "1.5.2" or later in the "1.5" branch.

Kai-Chen:
OK.  Should be much simpler.

To formalize this a bit:
* The file `.spark_version` will contain the required version as exactly 3 numbers separated by dot, e.g. `1.5.2`
* We will read the installed spark version and check it using the following rules:
  1. The first 2 (major and minor version) numbers must match exactly, and there should not be any non-digit characters
  1. Anything after the second dot is the patch version. We will take only the leading integer part, and this has to be greater or equal to the specified patch.

Corner cases for installed version (may not be realistic, but for completeness)
* 1.5
* 1.5-SNAPSHOT
* 1.5.-SNAPSHOT

Do we interpret all of the above as '1.5.0'?

AliTajeldin:
Let's not worry about the corner cases for now.  Spark versions so far have been consistent.

Kai-Chen:
How about, if the installed version doesn't have at least 3 numbers separated by dots, we will treat it as error and ask the user to install the required version?

AliTajeldin:
don't worry about corner cases.  We are not trying to create a generic version comparison function.  We are trying to figure out if the spark version is ok.  Just assume it has 3 parts.

Kai-Chen:
Removed the functions and test files, and updated `_env.sh` with the simpler check.

AliTajeldin:
nice and easy to understand.

AliTajeldin:
@Kai-Chen Travis CI test is failing.  It is not detecting the correct spark version. We get the following error:
```
Spark 1.5.2+hotdeploy detected. Please install Spark 1.5.2.
```

PR #1017, opened 2017-11-07
added discoverSchemaReturnString entrypoint to SmvApp

### LINKED ISSUE #1016: Enable functionality to retrieve a dataset's schema in json format from the python side
Currently, it is possible to retrieve/discover the schema by generating a schema file. The python interface should be able to get the schema of a dataset without having to create the file. Namely, there should be a function that when called, returns a json string representing the schema of a given dataset.

AntonioRecaldeRusso:
Fixes #1016 

Added entrypoint to smvapp.py, which returns a the schema string for a given .csv file.

### DIFF
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -136,13 +136,14 @@ class SmvApp(object):
     def config(self):
         return self.j_smvApp.smvConfig()
 
-    def discoverSmvSchema(self, path, csvAttributes, n=100000):
-        """Returns schema information of a csv file in json format
+    def discoverSchemaAsSmvSchema(self, path, csvAttributes, n=100000):
+        """Discovers the schema of a .csv file and returns a Scala SmvSchema instance
 
         path --- path to csvfile
-        n --- number of records used to discover schema
+        n --- number of records used to discover schema (optional)
+        csvAttributes --- Scala CsvAttributes instance (optional)
         """
-        return self._jvm.SmvPythonHelper.discoverSmvSchema(path, n, csvAttributes)
+        return self._jvm.SmvPythonHelper.discoverSchemaAsSmvSchema(path, n, csvAttributes)
 
     def inputDir(self):
         return self.config().inputDir()
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -136,6 +136,14 @@ class SmvApp(object):
     def config(self):
         return self.j_smvApp.smvConfig()
 
+    def discoverSchemaReturnString(self, path, n=100000):
+        """Returns json string containing the schema information of a csv file
+
+        path --- path to csvfile
+        n --- number of records used to discover schema
+        """
+        return self._jvm.SmvPythonHelper.discoverSchemaReturnString(path, n)
+
     def inputDir(self):
         return self.config().inputDir()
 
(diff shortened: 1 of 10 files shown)


src/main/scala/org/tresamigos/smv/python/SmvPythonProxy.scala
```diff
@@ -106,6 +106,10 @@ object SmvPythonHelper {
   def discoverSchema(path: String, nsamples: Int, csvattr: CsvAttributes): Unit =
     shell.discoverSchema(path, nsamples, csvattr)
 
+  def discoverSchemaReturnString(path: String, nsamples: Int): String = {
```
AliTajeldin: * The wording of the function name is odd.  how about `discoverSchemaAsString` (same on the python side)
* What about the csv attributes?  How does user specify if it has header, delimiter, etc.

src/main/scala/org/tresamigos/smv/python/SmvPythonProxy.scala
```diff
@@ -106,8 +106,9 @@ object SmvPythonHelper {
   def discoverSchema(path: String, nsamples: Int, csvattr: CsvAttributes): Unit =
     shell.discoverSchema(path, nsamples, csvattr)
 
-  def discoverSchemaReturnString(path: String, nsamples: Int): String = {
-    new SchemaDiscoveryHelper(SmvApp.app.sqlContext).discoverSchemaFromFile(path, nsamples).toString()
+  def discoverSmvSchema(path: String, nsamples: Int, csvattr: CsvAttributes): SmvSchema = {
```
AliTajeldin: The name of this function is very similar to the one above it but they do different things.  In the ideal situation we would rename both, but the above function (to file) is exposed to users and it would be a pain to change it's name at this point.  So let's rename this function to `discoverSchemaAsString` to highlight the fact that we are returning the discovered schema as a string.

src/main/python/smv/smvapp.py
```diff
@@ -136,6 +136,14 @@ def appName(self):
     def config(self):
         return self.j_smvApp.smvConfig()
 
+    def discoverSmvSchema(self, path, csvAttributes, n=100000):
+        """Returns schema information of a csv file in json format
```
AliTajeldin: not json

src/main/python/smv/smvapp.py
```diff
@@ -136,6 +136,14 @@ def appName(self):
     def config(self):
         return self.j_smvApp.smvConfig()
 
+    def discoverSmvSchema(self, path, csvAttributes, n=100000):
```
AliTajeldin: do we have a way to create an `CsvAttributes` instance on the python side?

src/main/scala/org/tresamigos/smv/python/SmvPythonProxy.scala
AntonioRecaldeRusso: The return value of `discoverSmvSchema` is a `SmvSchema` instance.
Should I rename it `discoverSchemaAsSmvSchema`?

src/main/python/smv/smvapp.py
AntonioRecaldeRusso: Yes. It is possible to create a `CsvAttributes` instance via the `_mkCsvAttr` method in SmvApp.
Should `csvAttributes` argument have a default value in this case?

AntonioRecaldeRusso:
Additionally, as part of this PR, the `discoverSchema` method will be renamed to `discoverSchemaToFile` wherever such a method discovers the schema and as a result creates a `.schema`
No renaming will be done wherever a `discoverSchema` method only discovers the schema (not resulting in a `.schema` file creation). E.g.: `discoverSchema` in `SchemaDiscoveryHelper`

PR #1021, opened 2017-11-12
I1020 filter abc

### LINKED ISSUE #1020: graph and ls do not work on project with ABC python SmvDatasets
The problem is that the `dataSetsForStage` method of the python datasetrepo will return the ABC as one valid dataset, but when `DataSetResolver` calls `findDataSetInRepo`, it will try to load the ABC, which fails.

ninjapapa:
Fixed #1020 

Python ABC SmvDataset's are filtered in `dataSetsForStage`.

### DIFF
--- a/src/main/python/smv/datasetrepo.py
+++ b/src/main/python/smv/datasetrepo.py
@@ -104,7 +104,8 @@ class DataSetRepo(object):
                             # Each package will contain among other things all of
                             # the modules that were imported into it, and we need
                             # to exclude these (so that we only count each module once)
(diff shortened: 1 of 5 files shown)


ninjapapa:
Sounds like python 3.5 doesn't event consider SmvDataset an Abstract class. 
Should use the following way to declare an ABC:

https://stackoverflow.com/questions/35673474/using-abc-abcmeta-in-a-way-it-is-compatible-both-with-python-2-7-and-python-3-5

AliTajeldin:
nice!

PR #1036, opened 2017-12-02
I1035 simpler smv time panel agg

### LINKED ISSUE #1035: Simplify smvTimePanelAgg implementation
`smvTimePanelAgg` method need to fill in the blank months (or other partial time) with null values. Current implementation calls `fillExpectedWithNull` (private[smv]) which calls `smvCollectSet`, which calls the UDAF `histstr`. 

The UDAF is quite memory-hungry especially when the key space and time dimension are large. If use `smvRePartition` to reduce the key space a round of unnecessary shuffling is need. In a client project, an aggregation on a 1.9B record data set takes about 2 min, the `smvRePartition` takes 8+ hours. 

Actually it is unnecessary for `smvTimePanelAgg` to call `smvCollectSet`. For the following case,

id | smvTime | v
---|-----|----
1 | M201201 | 1.0
1 | M201203 | 2.0
1 | M201203 | 3.1
1 | M201204 | 2.0

If the panel is from 201201 to 201204, we need to add a record `1, M201202, null`. To make sure that we only add that null record, current implementation has to use `smvCollectSet` to figure out that `M201202` is the only missing month for `id=1`. However actually since all Spark aggregation functions on the column `v` will ignore the `null` records, the aggregation result should be the same if we add all the following records,

id | smvTime |v 
---|----|---
1 | M201201 | null
1 | M201202 | null
1 | M201203 | null
1 | M201204 | null

In that case no need to call `smvCollectSet`, so no need to call the memory-hungry UDAF. On 1.6+ Spark, although we don't need to call the UDAF  (use `collect_set` function instead), the add-all-null is still simpler.

ninjapapa:
Fixed #1035


src/main/scala/org/tresamigos/smv/SmvGroupedData.scala
```diff
@@ -921,6 +800,56 @@ class SmvGroupedDataFunc(smvGD: SmvGroupedData) {
...
+   * 1, b, x
+   * 1, a, null
+   * 1, b, null
+   * 1, c, null
+   * }}}
+   **/
+  private[smv] def addExpectedWithNull(
+      colName: String,
+      expected: Set[String],
+      doFiltering: Boolean
```
AliTajeldin: What is `doFiltering`?  Please add a quick blurb about that in the comment above.

AliTajeldin:
@ninjapapa can you add the doc so we can merge this into master.

PR #1052, opened 2018-01-01
I1050 dynamic runtime config

### LINKED ISSUE #1050: Support for dynamic run time configuration

chengfei0913:
Fixes #1050 

Method `runModule` accepts `run_config` as dynamic runtime configuration. And the dynamic runtime configuration is always reset before every run operation.

### DIFF
--- a/src/main/python/scripts/smvserver.py
+++ b/src/main/python/scripts/smvserver.py
@@ -187,8 +187,8 @@ def getDatasetInstance(fqn):
     '''returns dataset object given a fqn'''
     return DataSetRepoFactory(SmvApp.getInstance()).createRepo().loadDataSet(fqn)
 
-def runModule(fqn):
-    '''runs module of given fqn'''
+def runModule(fqn, run_config=None):
+    '''runs module of given fqn and runtime configuration'''
     return SmvApp.getInstance().runModule("mod:{}".format(fqn))
 
 def getMetadataJson(fqn):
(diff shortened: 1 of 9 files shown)


src/main/scala/org/tresamigos/smv/SmvConfig.scala
```diff
@@ -239,14 +239,17 @@ class SmvConfig(cmdLineArgs: Seq[String]) {
...
 
+  // ---------- Dynamic Run Config Parameters key/values ----------
+  var dynamicRunConfig: Map[String, String] = Map.empty
+
   /** The FQN of configuration object for a particular run.  See github issue #319 */
   val runConfObj: Option[String] = cmdLine.runConfObj.get.orElse(mergedProps.get(RunConfObjKey))
 
   // ---------- User Run Config Parameters key/values ----------
-  val runConfigKeys: Seq[String] = splitProp("smv.config.keys")
+  lazy val runConfigKeys: Seq[String] = splitProp("smv.config.keys") ++ dynamicRunConfig.keySet
```
AliTajeldin: should probably make this a pure function.  If this set of keys changes from one run to the other, we will not see the new set of keys.

src/test/python/testDynamicRunConfig.py
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from test_support.smvbasetest import SmvBaseTest
+from smv import SmvApp
+
+class RunModuleWithRunConfigTest(SmvBaseTest):
```
AliTajeldin: nice test!

src/main/python/scripts/smvserver.py
```diff
@@ -200,14 +200,23 @@ def getMetadataJson(fqn):
...
+    function: run the module of given fqn and runtime configuration
     '''
     try:
         module_fqn = request.form['fqn'].encode("utf-8")
     except:
         raise err_res('MODULE_NOT_PROVIDED_ERR')
-    return ok_res(str(runModule(module_fqn)))
+    try:
+        encoded_run_config = None
+        run_config = request.form['run_config']
```
AliTajeldin: why do we need to use `request.form`?  Isn't it just a `json` object? Probably needs an explanation here as a comment.

chengfei0913:
Reopen for review

PR #1065, opened 2018-01-12
I1057 high order dqm

### LINKED ISSUE #1057: create higher order historical dqm
The current historical dqm interface is low level and forces users to define the `metadata` and `validateMetadata` for each module.  Client requested a higher order implementation where they can "mixi-in" a set of historical validators per module.

Proposed Interface Example:
```
@SmvHistoricalValidators(
  DistinctCountVariation(“id_count”, “id”, 0.1),
  AvgValueVariation(“amt_var”, “amount”, 0.2))
class MyModule(smv.SmvModule):
  def requiresDS(self): …
  def run(self, i): …
```

Users just add a decorator to an existing module to add historical validation checks.  In the above example, we added two validation rules.  Each rule takes a name parameter and validation specific parameters.  The first rule will check variation in distinct count of the “id” column.  The second rule checks for variation in average value of the “amount” column.  The third parameter (0.1 and 0.2) determine how much variation we allow.

It should be easy add additional validators.  It is similar to the low level metadata interface.  For example:
```
class NewValidator(smv.SmvHistoricalValidator):
  def __init__(self, name, arg1, arg2, arg3, …):
    pass
  def metadata(df):
    # compute value to track from current result df.
    val = ...
    return { name : val }
  def validateMetadata(cur_metadata, hist_metadata):
    # compare current metadata value to historical metadata values.
    return True/False
```

If the decorator interface does not work, an alternative is to mix-in a validation class:
```
class MyModule(smv.SmvModule, smv.SmvHistoricalValidators):
  def requiresDS(self): …
  def run(self, i): …
  def getValidators(self): return [
    DistinctCountVariation(“id_count”, “id”, 0.1),
    AvgValueVariation(“amt_var”, “amount”, 0.2)]
```

AliTajeldin:
Fixes #1057


src/main/python/smv/historical_validators.py
```diff
@@ -0,0 +1,70 @@
+def SmvHistoricalValidators(*validators):
+    """Decorator to specify set of validators to apply to an SmvDataSet.
+
+        Each validator should specify the standard `metadata` and `validateMetadata`
+        methods.  In addition, each validator will have a unique key defined that
+        will determine how it is stored in the final metadata structure.  This
+        decorator will take care of composing the union of all metadata from validators
+        and decomposing the full metatdata structure into individual pieces for each
```
kmannislands: metadata* instead of metatdata and I think you meant final* not fine in the hint

src/main/python/smv/historical_validators.py
```diff
@@ -0,0 +1,70 @@
...
+        cls.metadata = metadata
+        cls.validateMetadata = validateMetadata
+        return cls
+
+    return cls_wrapper
+
+class SmvHistoricalValidator(object):
+    """Base of all user defined historical validator rules.
+
+        Derived classes must override the `key`, `metadata`, and `validateMetadata`
```
kmannislands: It says derived classes must implement key themselves but it looks like there's an implementation here

AliTajeldin: good catch!

Issue #1060, opened 2018-01-11
Python API docs for v2.1.1.3 and v1.6.2.2 releases are broken 


laneb:
See e.g. [1.6.2.2 docs](http://tresamigossd.github.io/SMV/pythondocs/1.6.2.2/smv.html). None of the packages have any info. It's possible that this is the case for other releases as well.

laneb:
After regenerating the docs on the 1.6 branch, they aren't broken. It is not clear what the problem is/was. The last several 2.1 releases have been broken, but only 1.6.2.2 was broken (not 1.6.2.1).

laneb:
The problem was that the wrong Docker image was being used when building the docs #1064. The incorrect py4j version was causing `ImportErrors`.

laneb:
@AliTajeldin Should any of the existing documentation be fixed, or should we just wait for the next releases?

AliTajeldin:
No worries about old docs.  Just fix future releases should be fine.

laneb:
Will close this issue when we confirm that the next non-master release has correct docs.

laneb:
Docs at http://tresamigossd.github.io/SMV/pythondocs/1.6.2.3/smv.html are correct.

PR #1067, opened 2018-01-12
smv-release should use the previous docker image, not latest

### LINKED ISSUE #1064: smv-release should run sbt commands against previous Docker image, not latest
`smv-release` runs sbt commands, such as `sbt pytest` in docker containers. However, it enforces using the latest image (currently 1.5.2.17), which causes tests to fail when releasing other branches due to incorrect Spark version. Instead, the Docker image for the previous release (1.6.2.2 when releasing 1.6.2.3) should be used.

laneb:
Fixes #1064.


laneb:
The Python doc generation also uses the latest Docker image instead of the previous one. Appears to be the cause of #1060 as well. Will fix in this PR.

PR #1073, opened 2018-01-17
i1069 fix Py4J error in module discovery when file contains JavaObject

### LINKED ISSUE #1069: Runtime error when file in stage contains a Py4J JavaObject or PySpark Column
Maybe an atypical usage pattern, but it's come up in a client project. When trying to run the project, got an error whose gist is
```
py4j.Py4JException: Method fqn([]) does not exist
```
During module discovery, we are filtering the objects discovered in files first by attempting to get their `IsSmvDataSet` properties and then by comparing their fqns to the stage name. An assumption is made that if an object is not an `SmvDataSet`, it will not have property named `IsSmvDataSet` and we will get an attribute error before `fqn` is called on the object. However, Py4J has overridden how `JavaObjects` get attributes, and `IsSmvDataset` is resolving to a truthy `JavaMember`. Then we try to call a nonexistent `fqn` method of the `JavaObject`.

laneb:
Fixes #1069. Includes a small cleanup of the implementation of `datasetsForStage`.

### DIFF
--- a/src/main/python/smv/datasetrepo.py
+++ b/src/main/python/smv/datasetrepo.py
@@ -74,43 +74,47 @@ class DataSetRepo(object):
     getLoadDataSet = create_py4j_interface_method("getLoadDataSet", "loadDataSet")
 
 
-    def dataSetsForStage(self, stageName):
-        return self._moduleUrnsForStage(stageName, lambda obj: obj.IsSmvDataSet)
-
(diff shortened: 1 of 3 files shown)


AliTajeldin:
Really nice.  And thanks for adding the docs to the old code.  Really makes it much easier to understand.

PR #1079, opened 2018-01-18
I1078 Python API to receive SmvRunInfoCollector

### LINKED ISSUE #1078: Provide Python API for SmvRunInfoCollector and DqmValidationResult
Python code should be able to access `DqmValidationResult` and its members.

See umbrella ticket #1062

Kai-Chen:
Closes #1078

### DIFF
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -174,16 +174,22 @@ class SmvApp(object):
         """
         fqn = urn[urn.find(":")+1:]
         ds = self.repoFactory.createRepo().loadDataSet(fqn)
-        df = self.runModule(urn, forceRun, version)
-        # df, collector = self.runModule(urn, forceRun, version)
+        df, collector = self.runModule(urn, forceRun, version)
         return ds.df2result(df)
 
     def runModule(self, urn, forceRun = False, version = None, runConfig = None):
         """Runs either a Scala or a Python SmvModule by its Fully Qualified Name(fqn)
+
+        Use j_smvPyClient instead of j_smvApp directly so we don't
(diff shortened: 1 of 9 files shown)


src/main/python/smv/smvapp.py
```diff
@@ -174,16 +174,22 @@ def getModuleResult(self, urn, forceRun = False, version = None):
...
-        df = self.runModule(urn, forceRun, version)
-        # df, collector = self.runModule(urn, forceRun, version)
+        df, collector = self.runModule(urn, forceRun, version)
         return ds.df2result(df)
 
     def runModule(self, urn, forceRun = False, version = None, runConfig = None):
         """Runs either a Scala or a Python SmvModule by its Fully Qualified Name(fqn)
+
+        Use j_smvPyClient instead of j_smvApp directly so we don't
+        have to construct SmvRunCollector from the python side.
```
AliTajeldin: can you add a blurb about the return tuple as it took me a few seconds to figure out why there was a `[0]` after all the calls to this.

tools/conf/smv_pyshell_init.py.template
```diff
@@ -43,6 +43,32 @@ if os.path.exists(historyPath):
 
 atexit.register(save_history)
 
+# temporary: inspect validation result
```
AliTajeldin: should probably make this permanent and add it to `smvshell.py` (and document it as well).  Could be quite handy for shell users.

src/main/python/smv/smvapp.py
Kai-Chen: okiedo, coming up.

PR #1083, opened 2018-01-19
I1082 Collect metadata when a module is run

### LINKED ISSUE #1082: Collect metadata when running an SMV module
See #1062

Kai-Chen:
Close #1082 

## Requires PR #1079 and #1081

### DIFF
--- a/src/main/python/smv/runinfo.py
+++ b/src/main/python/smv/runinfo.py
@@ -1,0 +1,49 @@
+#
+# This file is licensed under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Easy Python access to SmvRunInfoCollector and related Scala classes.
+
+Todo:
+    * document example use
+"""
+
+import json
+
+class SmvRunInfoCollector(object):
+    """Python wrapper to its counterpart on the Scala side
+
+    Example:
+        df, coll = smvApp.runModule(...)
+        coll.dsFqns  # returns
+    """
+    def __init__(self, jcollector):
+        self.jcollector = jcollector
+
+    def fqns(self):
+        """Returns a list of FQNs for all datasets that ran"""
+        return self.jcollector.dsFqnsAsJava()
+
+    def dqm_validation(self, dsFqn):
+        """Returns the DQM validation result for a given dataset
+
+        Returns:
+            A dictionary representation of the dqm validation result
+
+        Raises:
+            py4j.protocol.Py4JError: if there is java call error or
+                there is no validation result for the specified dataset
+                (e.g. caused by a typo in the name)
+
+        """
+        java_result = self.jcollector.getDqmValidationResult(dsFqn)
+        return json.loads(java_result.toJSON())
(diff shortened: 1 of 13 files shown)


Kai-Chen:
Last commit also collects metadata history, per request from Jacob.

src/main/python/smv/runinfo.py
```diff
@@ -62,3 +62,18 @@ def metadata(self, dsFqn):
         """
         java_result = self.jcollector.getMetadata(dsFqn)
         return json.loads(java_result.toJson())
+
+    def metadata_history(self, dsFqn):
+        """Returns the metadata for a given dataset
+
+        Returns:
+            A dictionary representation of the metadata
```
laneb: Won't the result be a list of dictionaries?

Kai-Chen: That's correct :)
Another copy-and-paste error!

AliTajeldin: Copy-n-paste is EVIL!!!  😄

AliTajeldin:
LGTM.  awaiting approval from @jacobdr or @laneb as they will be the consumers of this.

src/test/python/testRunCmdLine.py
```diff
@@ -24,7 +24,7 @@ def smvAppInitArgs(cls):
 
     def test_can_run_module_from_cmdline(self):
         self.smvApp.j_smvApp.run()
-        a = self.smvApp.runModule(self.modUrn)
```
laneb: SmvBaseTest has a `df` method which you have already updated to take the `DataFrame` from the tuple returned by `runModule` . It would reduce noise a bit if we use that instead of indexing the result inline. Then again.

src/main/python/smv/smvshell.py
```diff
@@ -272,6 +273,18 @@ def _clear_from_sys_modules(names_to_clear):
                 sys.modules.pop(name)
                 break
 
+def show_run_info(collector):
```
laneb: How does a user in the shell get access to a `collector`?

Kai-Chen: It's returned as the second element in the tuple from `runModule`, as in
```
    res, coll = app.runModule('stage.modules.A')
```

src/main/python/smv/runinfo.py
Kai-Chen: Yes, but it saved me at least one minute on that one :)
Not to mention reducing significantly the probability of getting carpel-tunneled.

src/main/python/smv/runinfo.py
```diff
@@ -0,0 +1,79 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Easy Python access to SmvRunInfoCollector and related Scala classes.
+
+Todo:
```
laneb: Still todo?

src/main/python/smv/smvshell.py
laneb: But the app itself isn't really accessible from the shell. If there is a realistic use case for showing the run info in the shell, we need to make it simpler. Perhaps we should add a shell function that actually returns the collector with the `DataFrame`, or `df` should just have an option to display the run info at the time the `DataFrame` is returned.

Kai-Chen: I think in `smv-pyshell` we do provide the `app` instance, the same way that Spark provides sc and sqlContext.
It's done in the `pyshell_init` template file.

We can definitely modify the `df` method to have an option to print the run info.
Should I move this function there?

laneb: You're right, it is. I'm inclined to say it should not be exposed in that way, and we should not encourage users to use the `app` directly from the shell. I do think adding this output as an option of `df` would be preferable to the current approach, but I'm not set on it. @AliTajeldin can you weigh in?

laneb:
LGTM, except we should simplify showing the run info in the shell.

Issue #1092, opened 2018-01-26
Use json library to generate dependency graph json


AntonioRecaldeRusso:
In `SmvStageGraph.scala`, the createGraphJson method will generate an invalid json string if any of the metadata fields are a string with especial characters (e.g.: \n, \r, \t)

```scala
  def createGraphJSON(targetDSs: Seq[SmvDataSet] = Nil): String = {
    val g = new SmvDSGraph(app, stages, targetDSs)

    def toName(ds: SmvDataSet) = s"""\"${ds.fqn}\""""

    def toNodeStr(m: SmvDataSet) = {
      val dsType = m.dsType
      val nodeType = if (dsType == "Input") "file" else (
        // convert the first character of dsType to lowercase
        dsType.substring(0, 1).toLowerCase() + dsType.substring(1))
      s"""  {""" + "\n" +
      s"""    "fqn": ${toName(m)},""" + "\n" +
      s"""    "type": "${nodeType}",""" + "\n" +
      s"""    "version": ${m.version},""" + "\n" +
      s"""    "needsToRun": ${m.needsToRun},""" + "\n" +
      s"""    "description": "${m.description}"""" + "\n" +
      s"""  }"""
    }
```

ninjapapa:
This issue was fixed by #1235.

@AntonioRecaldeRusso close?

AntonioRecaldeRusso:
While the issue was fixed, the intention in this ticket as requested by @AliTajeldin at the time was that we use an actual json library to generate the dependency graph json. If that no longer applies, this ticket can be closed

AliTajeldin:
closing

PR #1096, opened 2018-01-27
dependency graph json string returns description metadata as defined in python module

### LINKED ISSUE #1095: Fix unescaped description metadata string in json graph
Quick fix for `createGraphJSON` generating an invalid json string in `SmvStageGraph.scala`.
Long term solution in #1092

AntonioRecaldeRusso:
Fixes #1095 

Previously, the `createGraphJson` method in `SmvStageGraph` used to return a hardcoded string and fqn  (`SMVModule {fqn}`) for python module's description metadata.
After this PR, the description metadata's of a python module is always displayed in the stage graph json string.

A more long term solution will be implemented in #1092


AliTajeldin:
nice!

PR #1102, opened 2018-01-30
#1100: allow glob pattern in file path of CSV files

### LINKED ISSUE #1100: handle case where csv file path contains * glob
It looks like newer version of HDFS do not allow `Path` to contain glob pattern in the `HDFS.getFileStatus()` method.  So if a CSV file had a glob pattern in the name, it will fail to get the modification time of the file.

In the short term, we should ignore file paths with glob pattern and in the long term ,we should expand the glob pattern and compute a hashed modification time based on all files that match the pattern (and rename the method to reflect the change).

AliTajeldin:
Fixes #1100

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SmvHDFS.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvHDFS.scala
@@ -109,13 +109,20 @@ private[smv] object SmvHDFS {
   }
 
   /**
-   * get modification time of a HDFS file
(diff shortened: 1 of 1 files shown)


laneb:
The intent is that any CSV file with a glob pattern is assumed unchanged, correct?

AliTajeldin:
correct.  In the future we may add support to do a hash of all glob matching csv files (but the name of the function would also have to change).

PR #1103, opened 2018-01-30
I1072 SmvSchema support for Byte and Short

### LINKED ISSUE #1072: Add SmvSchema support for ByteType
SMV doesn't know how to handle input data containing Bytes (or Hive tinyints).

laneb:
Resolves #1072.


AliTajeldin:
LGTM

AliTajeldin:
what is `dfFrom`?

laneb:
`dfFrom` is a method provided by the `SparkTestUtil` to do the equivalent of `SmvApp.createDF`. It actually just delegates to `createDF`. It does not persist a CSV, but it creates a `DataFrame` as though reading a CSV with the specified data. This means that an `SmvSchema` is constructed from the schema string and used to interpret the text data.

PR #1107, opened 2018-01-30
getRunInfo must follow links before getting fqn of dependency (#1106)

### LINKED ISSUE #1106: Error when getRunInfo of module that depends on link
`SmvModuleLink fqn should never be called`. Need to get the fqn on the _target_ of the link instead.

laneb:
Fixes #1106.


AliTajeldin:
@laneb I'm not seeing the actual fix.  All I see are changes to test code.  You may have forgotten to "git add" the required file.

laneb:
Yeah, that's what happened. I've amended to include the fix.

PR #1111, opened 2018-01-31
I1099 publish to hive method

### LINKED ISSUE #1099: Python SmvApp needs a method to publish module(s) to Hive programmatically
User writes a Python script that runs modules with a dynamically generated config, and wants to publish the results to Hive. Currently, only Scala `SmvApp` has a method for publishing to Hive, and it does not accept a run config argument.

laneb:
Resolves #1099. As a bonus, I fixed the pyshell function `exportToHive`, which was calling the Scala `exportToHive` with the wrong args. The Scala args had changed after the SmvRunInfoCollector was added. Also gave `exportToHive` a `runConfig` arg.

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SmvApp.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvApp.scala
@@ -373,6 +373,14 @@ class SmvApp(private val cmdLineArgs: Seq[String],
     val ds = dsm.inferDS(modName).head
     runDS(ds, forceRun, version, runConfig, collector=collector)
   }
+  
+  def publishModuleToHiveByName(modName: String,
+                                runConfig: Map[String, String],
+                                collector: SmvRunInfoCollector): Unit = {
+      val ds = dsm.inferDS(modName).head
+      setDynamicRunConfig(runConfig)
+      ds.exportToHive(collector)
+  }
 
   def getRunInfo(partialName: String): SmvRunInfoCollector =
     getRunInfo(dsm.inferDS(partialName).head)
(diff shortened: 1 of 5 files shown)


AliTajeldin:
I like the bonus 👍

src/main/scala/org/tresamigos/smv/SmvApp.scala
```diff
@@ -373,6 +373,14 @@ class SmvApp(private val cmdLineArgs: Seq[String],
     val ds = dsm.inferDS(modName).head
     runDS(ds, forceRun, version, runConfig, collector=collector)
   }
+  
+  def publishModuleToHiveByName(modName: String,
+                                runConfig: Map[String, String],
+                                collector: SmvRunInfoCollector): Unit = {
+      val ds = dsm.inferDS(modName).head
+      setDynamicRunConfig(runConfig)
```
AliTajeldin: I think the set dynamic run config should be fore the `inferDS` call as we may be adding support for updating SMV parameters in the runconfig as well.

laneb: The dynamic run config is set after `load` and `inferDS`  in `runModule` and `runModuleByName` as well. Should I change all 3 in this PR?

AliTajeldin: I think @kmannislands is taking care of the other two.  Might as well do it here to be consistent.

kmannislands: Yes, I have but this might go in first so I can just watch this carefully when rebasing

AliTajeldin: or merging 😄

src/main/scala/org/tresamigos/smv/SmvApp.scala
```diff
@@ -338,7 +338,7 @@ class SmvApp(private val cmdLineArgs: Seq[String],
         deletePersistedResults(Seq(ds))
 
       // set dynamic runtime configuration before run
-      setDynamicRunConfig(runConfig)
+      x(runConfig)
```
AliTajeldin: ????

laneb: Good question. That'll teach me to check my diffs and run my tests, though...

PR #1119, opened 2018-02-01
Smv Dynamic App Directory

### LINKED ISSUE #1084: Make all smv config dynamic
There is now a use case to have stages and dataDir at least dynamic.
Task is to extend the dynamic run config strategy previously employed to cover all smv config

kmannislands:
Fixes #1084 with a few other (hopefully) useful additions
- SMV's app directory can now be changed dynamically for python projects by calling `SmvApp.setAppDir('/foo/bar/')`. This will update all of the app's configuration properties as well as discoverable modules dynamically
- The running final props can now be inspected with ~~`SmvApp.mergedPropsJSON()`~~ `SmvApp.getCurrentProperties()` on py or scala sides or with simply `props()` from the shell

**Caveat:** Providing the `runConfig` param to `runModule` or `runModuleByName` has the side effect of setting the dynamic config not just for that run, but going forward. I would prefer one of two options:
  - `smvApp.setDynamicRunConfig` ought to be called separately by the caller of the run function first
 - Change our implementation of runConfig so that providing it as an arg to a `runModule` call only impacts that run, not any successive run

### DIFF
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -18,6 +18,7 @@ It is equivalent to ``SmvApp`` on Scala side
 import os
 import sys
 import traceback
+import json
 
 from py4j.java_gateway import java_import, JavaObject
 from pyspark import SparkContext
@@ -157,12 +158,13 @@ class SmvApp(object):
     def setDynamicRunConfig(self, runConfig):
         self.j_smvPyClient.setDynamicRunConfig(runConfig)
 
-    def mergedPropsJSON(self):
-        """JSON utf-8 string of current mergedProps as JSON:
+    def getCurrentProperties(self, raw = False):
+        """ Python dict of current megred props
             defaultProps ++ appConfProps ++ homeConfProps ++ usrConfProps ++ cmdLineProps ++ dynamicRunConfig
-            Where right wins out in map merge
+            Where right wins out in map merge. Pass optional raw param = True to get json string instead
         """
-        return self.j_smvApp.smvConfig().mergedPropsJSON()
+        merged_json = self.j_smvPyClient.mergedPropsJSON()
+        return merged_json if raw else json.loads(merged_json)
 
     def stages(self):
         """Stages is a function as they can be set dynamically on an SmvApp instance"""
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -94,7 +94,7 @@ class SmvApp(object):
 
         # AFTER app is available but BEFORE stages,
         # use the dynamically configured app dir to set the source path
-        self.prepend_source()
+        self.prepend_source(self.srcPathRel)
 
         self.stages = self.j_smvPyClient.stages()
 
@@ -142,6 +142,13 @@ class SmvApp(object):
 
     def config(self):
         return self.j_smvApp.smvConfig()
+
+    def mergedPropsJSON(self):
+        """JSON utf-8 string of current mergedProps:
+            defaultProps ++ appConfProps ++ homeConfProps ++ usrConfProps ++ cmdLineProps ++ dynamicRunConfig
+            Where right wins out in map merge
+        """
+        return self.j_smvApp.smvConfig().mergedPropsJSON()
 
     def appId(self):
         return self.config().appId()
@@ -201,6 +208,8 @@ class SmvApp(object):
             - SmvRunInfoCollector contains additional information
               about the run, such as validation results.
         """
+        # This allows discovery of py mods for dynamic appDirs passed on RC
+        self.prepend_source_if_needed(runConfig)
         java_result = self.j_smvPyClient.runModule(urn, forceRun, self.scalaOption(version), runConfig)
         return (DataFrame(java_result.df(), self.sqlContext),
                 SmvRunInfoCollector(java_result.collector()) )
@@ -216,6 +225,8 @@ class SmvApp(object):
             - SmvRunInfoCollector contains additional information
               about the run, such as validation results.
         """
+        # This allows discovery of py mods for dynamic appDirs passed on RC
+        self.prepend_source_if_needed(runConfig)
         java_result = self.j_smvPyClient.runModuleByName(name, forceRun, self.scalaOption(version), runConfig)
         return (DataFrame(java_result.df(), self.sqlContext),
                 SmvRunInfoCollector(java_result.collector()) )
@@ -330,10 +341,16 @@ class SmvApp(object):
     def defaultTsvWithHeader(self):
         return self._mkCsvAttr(delimier='\t', hasHeader=True)
 
-    def prepend_source(self,):
-        smvAppDir = self.j_smvApp.smvConfig().appConfPath()
-        codePath = os.path.abspath(os.path.join(smvAppDir, self.srcPathRel))
+    def prepend_source_if_needed(self, runConfig):
+        """ Changes the python path if appDir "magic key" is included on rc """
+        if(runConfig["smv.app.dir"]):
+            self.prepend_source(self.srcPathRel, runConfig["smv.app.dir"])
 
+    def prepend_source(self, relPath, smvAppDir = None):
+        if (not smvAppDir):
+            smvAppDir = self.j_smvApp.smvConfig().appConfPath()
+        codePath = os.path.abspath(os.path.join(smvAppDir, relPath))
+        print "Prepending code path: " + codePath
         # Source must be added to front of path to make sure it is found first
         sys.path.insert(1, codePath)
 
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -49,7 +49,8 @@ class SmvApp(object):
     # default rel path for python sources from appDir
     srcPathRel = "src/main/python"
 
-    # keep track of the last appDir that was explicitly set
+    # keep track of the last appDir that was explicitly set so we can remove it
+    # from the sys.path when another one is to avoid cluttering the path
     lastCodePath = None
 
     @classmethod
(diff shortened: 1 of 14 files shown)


src/main/scala/org/tresamigos/smv/SmvConfig.scala
```diff
@@ -200,13 +213,41 @@ class SmvConfig(cmdLineArgs: Seq[String]) {
...
-  // merge order is important here.  Highest priority comes last as it will override all previous
-  private[smv] val mergedProps = defaultProps ++ appConfProps ++ homeConfProps ++ usrConfProps ++ cmdLineProps
+  // ---------- Dynamic Run Config Parameters key/values ----------
+  var dynamicRunConfig: Map[String, String] = Map.empty
+
+  // computeMergedProps() is used as a cache to avoid expensive IO if appDir is unchanged
+  private[smv] def mergedProps = { computeMergedProps(doIO = false) ++ dynamicRunConfig }
+
+  // return merged props as json so its easier on the py, used to expose merged props
+  def mergedPropsJSON = {
```
Kai-Chen: May want to consider moving this function to the `SmvPythonClient` since json serialization usually means, as it does here, it's going to go down a network socket.
And within the JVM, code can deal with the `mergedProps` directly.

src/main/scala/org/tresamigos/smv/python/SmvPythonProxy.scala
```diff
@@ -237,6 +237,18 @@ class SmvPyClient(val j_smvApp: SmvApp) {
     }
   }
 
+  def setDynamicRunConfig(runConfig: java.util.Map[String, String]): Unit = {
+    val dynamicRunConfig: Map[String, String] = if (null == runConfig) Map.empty else mapAsScalaMap(runConfig).toMap
+    j_smvApp.smvConfig.dynamicRunConfig = dynamicRunConfig
+  }
+
+  def setAppDir(appDir: String): Unit = {
```
kmannislands: Have the feeling these assignments shouldn't be in the proxy but otherwise the caller will have to say something like: `self.j_smvApp.smvConf().appDir_(appDir)` which seems less clear

src/test/python/testDynamicConfigAppDir.py
```diff
@@ -0,0 +1,117 @@
...
+            a child of the cwd, and then that the module is avaiable and can be run once appDir is set to the project dir """
+        # make sure running the module without setting the app dir fails and raises
+        err = None
+        try:
+            self.smvApp.runModule(self.modUrn)[0]
+        except Exception as e:
+            err = e
+
+        self.assertEqual(isinstance(err, Exception), True)
+
```
Kai-Chen: `self.assertRaises(Py4jException, lambda: self.smvApp.runModule(...))`
is the idiomatic expression.

src/main/python/smv/smvapp.py
```diff
@@ -138,6 +145,29 @@ def appName(self):
...
+        self.j_smvPyClient.setAppDir(appDir)
+        # this call will use the dynamic appDir that we just set ^
+        # to change sys.path, allowing py modules to be discovered by python
+        self.prepend_source(self.srcPathRel)
+
+    def setDynamicRunConfig(self, runConfig):
+        self.j_smvPyClient.setDynamicRunConfig(runConfig)
+
+    def mergedPropsJSON(self):
+        """JSON utf-8 string of current mergedProps as JSON:
```
Kai-Chen: May want to consider returning a python dictionary and rename the method simply `appProperties` or something similar.

Most callers can't work directly with the json string.
So it's a better api to give the caller what's more useful -- a dictionary works well here.

src/main/scala/org/tresamigos/smv/python/SmvPythonProxy.scala
Kai-Chen: I think the python-to-java calls should go through the proxy as much as possible.
This way we can handle data type conversion very consistently and leave the rest of the application out of that.
Also we easily tailor any minor differences in the Python and Scala APIs.

The going-straight-to-scala through daisy-chains is more of an escape hatch.

src/test/python/testDynamicConfigAppDir.py
```diff
@@ -0,0 +1,111 @@
...
+    modName = 'modules.A'
+
+    # project B stuff
+    projBDir = 'project-b/'
+    diffStageUrn = 'mod:notstage.notmodules.NotModuleA'
+    # same stage/name as project_a modA
+    sameStageUrn = 'mod:stage.modules.A'
+
+    def setUp(self):
+        """ Setup the stuff needed for all test cases ONCE """
```
laneb: If you want it to run just once before all the tests then you want `setUpClass`.

src/test/python/testDynamicConfigAppDir.py
```diff
@@ -0,0 +1,111 @@
...
+        """ Reset the app dir to what it was initially after EACH test. Should be '.' '"""
+        self.smvApp.setAppDir(self.initial_app_dir)
+
+    @classmethod
+    def tearDownClass(cls):
+        # run default class teardown
+        super(RunModuleWithDynamicConfigAppDirTest, cls).tearDownClass()
+
+    def test_mods_available_to_run(self):
+        """ The intention of this test is to verify that the module is initially not avaiable because the project dir is
```
laneb: Can you shorten this docstring a bit? They get printed with the test name and result when the test runs, so briefer is better.

src/test/python/testDynamicConfigAppDir.py
```diff
@@ -0,0 +1,111 @@
...
+            a child of the cwd, and then that the module is avaiable and can be run once appDir is set to the project dir """
+        # make sure running the module without setting the app dir fails and raises
+        self.assertRaises(Exception, lambda: self.smvApp.runModule(self.modUrn))
+
+        # set the app dir
+        self.smvApp.setAppDir(self.proj_a_path)
+
+        # Now try to run the module. This time we expect success
+        A = self.smvApp.runModule(self.modUrn)[0]
```
laneb: In our tests in general, I would prefer if we used `self.df(self.modUrn) over `self.smvApp.runModule(self.modUrn)[0]`. It's easier to read.

src/test/python/testDynamicConfigAppDir.py
```diff
@@ -0,0 +1,111 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from test_support.smvbasetest import SmvBaseTest
+from smv import SmvApp
+import json
+import sys
+
+# TODO figure out why this kills all the successive tests
```
laneb: Still TODO?

src/main/scala/org/tresamigos/smv/python/SmvPythonProxy.scala
laneb: I tend to prefer the daisy chains, because there's one less file/entity to trace the call through. I agree the proxy makes things clearer in this case, though.

src/main/scala/org/tresamigos/smv/SmvConfig.scala
```diff
@@ -200,13 +211,36 @@ class SmvConfig(cmdLineArgs: Seq[String]) {
     "smv.class_dir"   -> "./target/classes"
   )
 
-  // merge order is important here.  Highest priority comes last as it will override all previous
-  private[smv] val mergedProps = defaultProps ++ appConfProps ++ homeConfProps ++ usrConfProps ++ cmdLineProps
+  // ---------- Dynamic Run Config Parameters key/values ----------
+  var dynamicRunConfig: Map[String, String] = Map.empty
+
+  // computeMergedProps() is used as a cache to avoid expensive IO if appDir is unchanged
+  private[smv] def mergedProps = { computeMergedProps(doIO = false) ++ dynamicRunConfig }
```
laneb: Isn't this just `mergedPropsCache ++ dynamicRunConfig`?

src/main/scala/org/tresamigos/smv/SmvConfig.scala
```diff
@@ -200,13 +211,36 @@ class SmvConfig(cmdLineArgs: Seq[String]) {
...
+
+  // initially, the merged props cache should do the io to read app conf
+  private var mergedPropsCache: Map[String,String] = ioBasedConf
+
+  /**
+   * This function is implemented as a stupid-simple cache to avoid doing IO to obtain props every time
+   */
+  private def computeMergedProps(doIO: Boolean = false): Map[String,String] = {
```
laneb: I think `computeMergedProps` is unnecessary. The cache update can just be moved into `appDir_=` and `mergedProps` can just go to `mergedPropsCache` directly.

src/main/scala/org/tresamigos/smv/SmvConfig.scala
```diff
@@ -200,13 +211,36 @@ class SmvConfig(cmdLineArgs: Seq[String]) {
...
+    if (!doIO) return mergedPropsCache;
+    // make the IO happen again
+    mergedPropsCache = ioBasedConf
+    mergedPropsCache
+  }
+
+  // calling this performs the file io, hence the name
+  private def ioBasedConf: Map[String, String] = {
```
laneb: Can we call this something more specific, like `loadConfFromFile`? "ioBased" is pretty vague and there's no "o" involved.

src/main/scala/org/tresamigos/smv/SmvConfig.scala
```diff
@@ -188,9 +188,20 @@ class SmvConfig(cmdLineArgs: Seq[String]) {
...
-  private val appConfProps  = _loadProps(pathJoin(cmdLine.smvAppDir(), cmdLine.smvAppConfFile()))
-  private val usrConfProps  = _loadProps(pathJoin(cmdLine.smvAppDir(), cmdLine.smvUserConfFile()))
-  private val homeConfProps = _loadProps(DEFAULT_SMV_HOME_CONF_FILE)
+  // default the app dir to the cmdLine value.
+  private var _appDir: String = cmdLine.smvAppDir()
+
+  // getter and setter because we need to recompute merged props when appDir is set explicitly
+  def appDir : String = _appDir
+
+  def appDir_= (appDirPath: String): Unit = {
```
laneb: Opinion: changing the semantics of `=` to implicitly read from a file and update a cached value is a good way to confuse future devs. It would be a lot clearer to just have a `setAppDir` method.

src/test/python/testDynamicConfigAppDir.py
kmannislands: Fortunately not 😅

PR #1127, opened 2018-02-03
I1126 - Update --driver-class-path to also supply a relative path, add support for `=` options when merging props

### LINKED ISSUE #1126: User-supplied driver class path not being merged with smv-* tool commands
We seem to treat the `--driver-class-path` different than we do other Spark arguments passed from the caller -- see [here](https://github.com/TresAmigosSD/SMV/blob/master/tools/_pyenv.sh#L17). Should merge this like we do the other options.

jacobdr:
Greater flexibility for supplying --jars, and support for merging --driver-class-path spark argument

Before we did not handle the case of when --option=abc with an equals sign. Additionally, when we ran in cluster mode, the absolute path supplied to --driver-class-path did not work since the JAR is placed by Spark at the root of the YARN container, and therefore needs to be referenced by relative path.

Closes #1126

### DIFF
--- a/tools/_env.sh
+++ b/tools/_env.sh
@@ -32,17 +32,33 @@ function split_smv_spark_args()
 
     # Need to extract the --jars option so we can concatenate those jars with
     # the APP_JAR when we run the spark-submit. Spark will not accept 2 separate
-    # --jars options
+    # --jars options. Also need to handle the case when user uses equal signs (--jars=xyz.jar)
+    # Same for --driver-class-path
     while [ $# -ne 0 ]; do
         if [ "$1" == "--jars" ]; then
             shift
             EXTRA_JARS="$1"
+            shift
+        # See http://wiki.bash-hackers.org/syntax/pe#search_and_replace for bash string parsing
+        # tricks
+        elif [ ${1%%=*} == "--jars" ]; then
+            local ACTUAL_JARS_PORTION="${1#*=}"
+            EXTRA_JARS="${ACTUAL_JARS_PORTION}"
+            # Only need to shift once since we dont have a space separator
+            shift
+        elif [ "$1" == "--driver-class-path" ]; then
+            EXTRA_DRIVER_CLASSPATHS="$1"
+        elif [ ${1%%=*} == "--driver-class-path" ]; then
+            local ACTUAL_CLASSPATHS_PORTION="${1#*=}"
+            EXTRA_DRIVER_CLASSPATHS="${ACTUAL_CLASSPATHS_PORTION}"
+            # Only need to shift once since we dont have a space separator
             shift
         else
           SPARK_ARGS=("${SPARK_ARGS[@]}" "$1")
           shift
         fi
     done
+
 }
 
 function find_file_in_dir()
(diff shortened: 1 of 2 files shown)


jacobdr:
1 test failed, but seems very unrelated to my change.... https://travis-ci.org/TresAmigosSD/SMV/builds/336856550#L2481

tools/_env.sh
```diff
@@ -32,17 +32,33 @@ function split_smv_spark_args()
...
     # the APP_JAR when we run the spark-submit. Spark will not accept 2 separate
-    # --jars options
+    # --jars options. Also need to handle the case when user uses equal signs (--jars=xyz.jar)
+    # Same for --driver-class-path
     while [ $# -ne 0 ]; do
         if [ "$1" == "--jars" ]; then
             shift
             EXTRA_JARS="$1"
             shift
+        # See http://wiki.bash-hackers.org/syntax/pe#search_and_replace for bash string parsing
```
AliTajeldin: let's keep this simple.  Only support `--jars x` and `--driver-class-apth y`.  No need to support both methods of supplying parameters.  Lets code to grok and maintain.

tools/_pyenv.sh
```diff
@@ -13,6 +13,12 @@ export PYTHONPATH="$SMV_HOME/src/main/python:$PYTHONPATH"
...
 export SPARK_PRINT_LAUNCH_COMMAND=1
 
+# We eagerly add the basename of the APP_JAR so that this same command
+# works in YARN Cluster mode, where the JAR gets added to the root directory
+# of the container. Also note the colon separator
 function run_pyspark_with () {
-  "${SMV_SPARK_SUBMIT_FULLPATH}" "${SPARK_ARGS[@]}" --jars "$APP_JAR,$EXTRA_JARS" --driver-class-path "$APP_JAR" $1 "${SMV_ARGS[@]}"
+  "${SMV_SPARK_SUBMIT_FULLPATH}" "${SPARK_ARGS[@]}" \
+    --jars "$APP_JAR,$EXTRA_JARS" \
+    --driver-class-path "$APP_JAR:$(basename ${APP_JAR}):$EXTRA_DRIVER_CLASSPATHS" \
```
AliTajeldin: out of curiosity, what happens with `EXTRA_DRIVER_CLASSPATHS` is empty?  Does spark `--driver-class-path` handle the case with a trailing `:`?

tools/_env.sh
jacobdr: I disagree... If its valid Spark syntax, I feel like we should support it. The errors were really wacky and I had to spend some time with Lane to understand this.. And he said he only knew about it because he ran into the bug some time ago.

tools/_pyenv.sh
jacobdr: Not that I really trust [this SO page](https://stackoverflow.com/questions/14979328/why-does-my-classpath-only-work-with-a-colon-on-the-end), but I have validated that it seems to assume it to mean the current directory (so it becomes like `file1.jar:.`)

tools/_env.sh
AliTajeldin: ok, if multiple people ran into this will allow the parsing duplication.

Issue #989, opened 2017-09-28
Use log4j instead of console log messages in SMV.


yw-yang:
When running smv jobs through terminal, lots of information will be printed on console but won't be stored on server. In most cases, an external log file is pretty useful for the ease of debugging and future reference.

Could SMV store console log to one external file like smv.log automatically?

ninjapapa:
Using UNIX tee command and pipe should be able to cover this use case:
```shell
smv-pyrun ..... 2>&1 |tee smv.log
```

As an action item, add this to the user doc.

yw-yang:
@ninjapapa Yes, users can use "tee" command to cover this issue but making it as an automatic behavior still has additional benefits:
1. User always prefers shorter command
2. Log management should be clear with the growth of project

Shall we add "smv.logDir" as a config parameter and by default, it points to current path?

AliTajeldin:
We should utilize log4j for output logging from within SMV.  The user can then decide how to redirect that output (console, file, logstash, etc).  Applications depending on SMV can then configure the log4j properties of SMV explicitly.

Changing title of ticket to reflect need for log4j logging.

ninjapapa:
@AliTajeldin will that solve the problem of python side logging messages?

AliTajeldin:
yes.  Logging should be in the SMV core with python methods exposed for python SMV code to utilize.  This way, both Scala and python code would log to the same place and with a single logger configuration.  And since we are using log4j, the same logger config can be used for both spark and SMV.

laneb:
It would also be valuable to have debug logs available in the logs. E.g., 

* What objects are inspected when searching for SmvModules
* What module is being run and when a 'subtask' of running a module starts (dqm, metadata, persisting, etc.) 
* What metadata is generated for a module

jacobdr:
+1

Need to ale sure we are explicit to users about how they can change the levels, and add logging to their own applications.

The Java -D options might look funny to some of them, so I feel documentation will be especially important here.

laneb:
Also: 
* **Warn** when fail to import a stage
* **Warn** about errors suppressed by walk_packages
* **Debug** listing the `sys.path` after a `DataSetRep` is constructed

laneb:
I've added the logs, but I'm having trouble getting the logs to be printed to the console. 
* log4j only looks for configuration in Spark's conf dir
* the rootLogger configuration doesn't get inherited by SMV's logger

[...]

PR #1141, opened 2018-02-19
move logic for resetting sys.path out ofSmvApp.prepend_source (#1139)

### LINKED ISSUE #1139: smv-pyshell initializes with incorrect sys.path
The recent dynamic run config work changed `SmvApp.prepend_path` to remove previously added paths from the `sys.path` before adding the new path. In `smv-pyshell-init.py.template`, we use `prepend_source` to add the `conf` dir to the path, then import everything from `conf/smv_shell_app_init.py`. This occurs after `src/main/python` is added to the path, causing `src/main/python` to be dropped from the path before the user is dropped into the shell. As a result users are unable to run any of their modules.

laneb:
Fixes #1139.

### DIFF
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -49,11 +49,7 @@ class SmvApp(object):
     _instance = None
 
     # default rel path for python sources from appDir
-    srcPathRel = "src/main/python"
-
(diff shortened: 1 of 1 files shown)


laneb:
I'll include that in #989.

PR #1142, opened 2018-02-19
I1137 run info run config

### LINKED ISSUE #1137: get_run_info should accept runConfig argument
To retrieve a module's run info when it was run with a particular config, the config will need to be the same at the time of requesting the info. This is because metadata and validation is versioned by hash. Users should be able to provide the `runConfig` they used when running a module at the time they request the run info.

laneb:
Resolves #1137. Needed to pull in the fix from #1139 to test this. Merge that one first to see a more helpful diff.


laneb:
GitHub hasn't updated the diff. Will replace with a new PR.

PR #1145, opened 2018-02-20
Fix datasetrepos ability to clear mods for multipart stages

### LINKED ISSUE #1144: clear_sys_modules is broken for multipart stages
It appears that the `clear_sys_modules` method on python dataset repos doesn't actually successfully clear loaded modules for multi-part stages.

kmannislands:
Fixes #1144

My approach is to assemble a list of stages and stage stubs that need to be removed and then iterate all of the loaded modules and to use the existing criteria to try to remove the fqn from `sys.modules`.

Also, I extensively renamed stuff to (hopefully) make more sense.

### DIFF
--- a/src/main/python/smv/datasetrepo.py
+++ b/src/main/python/smv/datasetrepo.py
@@ -43,27 +43,21 @@ class DataSetRepo(object):
         self._clear_sys_modules()
 
     def _clear_sys_modules(self):
-        """Clear all client modules from sys.modules
         """
-        # First, we assemble a list of fqn stubs we want to clear from the sys.modules
-        # If stage name is like appname.stagename, we want list('appname', 'appname.stagename')
+            Clear all client modules from sys.modules
+            If modules have names like 'stage1.stage2.file.mod', then we have to clear all of
+            set( 'stage1', 'stage1.stage2', 'stage1.stage2.file', 'stage1.stage2.file.mod' )
+            from the sys.modules dictionary to avoid getting cached modules from python when
+            we contruct a new DSR.
+        """
+        # { 'stage1' } from our example
+        fqn_stubs_to_remove = { fqn.split('.')[0] for fqn in self.smvApp.stages() }
 
-        # this adds the full stage names to the list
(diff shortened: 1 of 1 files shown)


src/main/python/smv/datasetrepo.py
```diff
@@ -45,10 +45,24 @@ def __init__(self, smvApp):
...
         """
-        for fqn in list(sys.modules.keys()):
-            for stage_name in self.smvApp.stages():
-                if fqn == stage_name or fqn.startswith(stage_name + "."):
-                    sys.modules.pop(fqn)
+        # First, we assemble a list of fqn stubs we want to clear from the sys.modules
+        # If stage name is like appname.stagename, we want list('appname', 'appname.stagename')
+
+        # this adds the full stage names to the list
+        fqn_stubs_to_remove = list(self.smvApp.stages())
```
laneb: We can simplify this method significantly. Its effect is to remove every subpackage of the stub of every stage. Removing packages starting with `airline`, `airline.etl`, or`airline.category`, is the same as removing packages starting `airline`.  Also `fqn_stubs_to_remove` can be kept unique by using a set instead of a list.

So, I think 52 - 59 can be reduced to
```
fqn_stubs_to_remove = {fqn.split('.')[0] for fqn in self.smvApp.stages()}
```

Since we are now removing every subpackage of every stub in `fqn_stubs_to_remove`, 62 - 66 can be simplified by just checking if the prefix of `loaded_mod_fqn` is in `fqn_stubs_to_remove`:
```
for stage_name in self.smvApp.stages():
  if stage_name.split('.')[0] in fqn_stubs_to_remove:
    sys.modules.pop(stage_name)
```

If you want to be functional
```
stub= lambda fqn: fqn.split('.')[0]
fqn_stubs_to_remove = {stub(fqn) for fqn in self.smvApp.stages()}
sys.modules = [fqn for fqn sys.modules if stub(fqn) not in fqn_stubs_to_remove ]
```

AliTajeldin:
Thanks for the extra description.  Should help the next person understand what is going on.

PR #1147, opened 2018-02-21
I989 better logging

### LINKED ISSUE #989: Use log4j instead of console log messages in SMV.

laneb:
I don't think this fully resolves #989, because I haven't figured out why the smv logger doesn't seem to inherit the root loggers level. I'd like to merge in this logging in the meantime. 

Note that I've moved the tools for package introspection from `smv.utils` into the `DataSetRepo` class, because that was the only place they were used and logging from them was easier this way.

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SmvDataSet.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvDataSet.scala
@@ -181,7 +181,7 @@ abstract class SmvDataSet extends FilenamePart {
       queries foreach {app.sqlContext.sql(_)}
     }
 
-    logAction(f"PUBLISHING ${fqn} TO HIVE: ${queries.mkString(";")}", _export)
+    doAction(f"PUBLISHING ${fqn} TO HIVE: ${queries.mkString(";")}", _export)
 
   }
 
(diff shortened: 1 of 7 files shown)


src/main/scala/org/tresamigos/smv/SmvDataSet.scala
```diff
@@ -307,28 +314,33 @@ abstract class SmvDataSet extends FilenamePart {
                attr: CsvAttributes = CsvAttributes.defaultCsv): DataFrame =
     new FileIOHandler(app.sqlContext, path).csvFileWithSchema(attr)
 
+  /**
+   * Perform an action and log the amount of time it took
+   */
+  def doAction(desc: String, action: () => Unit): Unit = {
```
AliTajeldin: nice!

AliTajeldin:
This is fantastic.  Please open a separate ticket to resolve the root logger issue.

PR #1151, opened 2018-02-22
enable SmvRunInfoCollector methods to accept partial name for a dataset (#1149)

### LINKED ISSUE #1149: SmvRunInfoCollector methods should accept partial names
Currently its methods like `metadata` require the FQN of the target module.

laneb:
Resolves #1149.

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SmvRunInfoCollector.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvRunInfoCollector.scala
@@ -50,21 +50,38 @@ class SmvRunInfoCollector {
    *
    * @throws NoSuchElementException if the dataset is not known to the collector
    */
-  def getDqmValidationResult(dsFqn: String): DqmValidationResult = runInfo(dsFqn).validation
+  def getDqmValidationResult(dsName: String): DqmValidationResult = inferRunInfo(dsName).validation
 
   /**
    * Returns the metadata for a given dataset
    *
    * @throws NoSuchElementException if the dataset is not known to the collector
    */
-  def getMetadata(dsFqn: String): SmvMetadata = runInfo(dsFqn).metadata
+  def getMetadata(dsName: String): SmvMetadata = inferRunInfo(dsName).metadata
 
   /**
    * Returns the metadata history for a given dataset
    *
    * @throws NoSuchElementException if the dataset is not known to the collector
    */
-  def getMetadataHistory(dsFqn: String): SmvMetadataHistory = runInfo(dsFqn).metadataHistory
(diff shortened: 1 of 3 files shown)


src/main/scala/org/tresamigos/smv/SmvRunInfoCollector.scala
```diff
@@ -50,21 +50,38 @@ class SmvRunInfoCollector {
...
+   * @throws SmvRuntimeException if no dataset or multiple dataset fqns ends with dsName
+   */
+  def inferRunInfo(dsName: String): SmvRunInfo = {
+    val candidates = dsFqns filter (_.endsWith(dsName))
+    val fqn = candidates.size match {
+      case 0 =>
+        throw new SmvRuntimeException(s"""No run info collected for module named [${dsName}]""")
+      case 1 => candidates.head
+      case _ =>
+        throw new SmvRuntimeException(s"[${dsName}] could refer to [${candidates.mkString(", ")}]")
```
cloudysail: I thought we were going to support both `FQN` and `lastName`. 


If we only support `lastName`, then we need to adjust all our tests and all the places where the API is invoked. Would that be too much? Also, we allow different FQN but the same `lastName`(e.g, A.B.test and C.D.test), right? Then we won't support such cases if only `lastName` supported.

jacobdr:
I think providing a.b.test and c.d.test throws an exception as not providing a specific enough target, and has a very good error message about the ambiguity and possible options.

For clarity not sure what last names mean — I would say this situation is where two modules have the same name but different FQNs.

The “full” FQN should be resolved without issue so I think this is not a breaking change.

Most important, after working with the run info collector quite a bit now, this is just the first of several changes we probably need to make in order to make it easier to work with.

Jacob Roberts

laneb:
If you have a `SmvRunInfoCollector` called `info` with run info for `a.b.test` and `c.d.test, 
`info.metadata("test")` will fail because it is ambiguous. Meanwhile, `info.metadata("a.b.test")` `info.metadata("c.d..test")` will give you run info for `a.b.test` and `c.d.test`, respectively. This is the same behavior as other methods that infer a dataset based on a partial name, like `df`. No need to rewrite any tests - you can see from the CI that all the tests are passing except for a spurious `ExecutorLostFailure`.

cloudysail:
Good to know that. I think we can close this PR since all the changes are included in #1152.

laneb:
I've opened separate PRs for these fixes in order to maintain the one-to-one relationship between issue, branch, and PR. Not a big issue if you'd rather accept it all in one merge request, but IMO keeping them separate makes the history a little clearer.

PR #1152, opened 2018-02-22
I1150 run info show report for single dataset

### LINKED ISSUE #1150: SmvRunInfoCollector show_report is too detailed
Calling `show_report` routinely causes Jupyter to complain that the result is too large to display. We need to reduce the size/detail of the report to make it readable. It may be enough to let the user specify an a dataset to show a report for.

laneb:
Resolves #1150.

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SmvRunInfoCollector.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvRunInfoCollector.scala
@@ -71,17 +71,23 @@ class SmvRunInfoCollector {
    *
    * @throws SmvRuntimeException if no dataset or multiple dataset fqns ends with dsName
    */
-  def inferRunInfo(dsName: String): SmvRunInfo = {
+  def inferFqn(dsName: String): String = {
     val candidates = dsFqns filter (_.endsWith(dsName))
-    val fqn = candidates.size match {
+    candidates.size match {
       case 0 =>
         throw new SmvRuntimeException(s"""No run info collected for module named [${dsName}]""")
       case 1 => candidates.head
       case _ =>
         throw new SmvRuntimeException(s"[${dsName}] could refer to [${candidates.mkString(", ")}]")
     }
-    runInfo(fqn)
   }
+
+  /**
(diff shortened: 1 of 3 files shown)


cloudysail:
Seem mixed with #1151. Is this intentional?

laneb:
I needed work from `i1149_run_info_collector_partial_name` in order to implement this, so I branched from there instead of from `master`. If it would be helpful, I can close this PR and reopen after the other branch is merged so 50b5a26 isn't shown in the diff.

cloudysail:
Aha, then this is OK

src/main/scala/org/tresamigos/smv/SmvRunInfoCollector.scala
```diff
@@ -50,21 +50,44 @@ class SmvRunInfoCollector {
...
+   * Returns the unique fqn ending with @param dsName from the datasets collected
+   *
+   * @throws SmvRuntimeException if no dataset or multiple dataset fqns ends with dsName
+   */
+  def inferFqn(dsName: String): String = {
+    val candidates = dsFqns filter (_.endsWith(dsName))
+    candidates.size match {
+      case 0 =>
+        throw new SmvRuntimeException(s"""No run info collected for module named [${dsName}]""")
+      case 1 => candidates.head
```
cloudysail: So `dsName` here could be either the FQN or last part of the FQN, right?
For example, if FQN is `a.b.test`, then the value of parameter `dsName` passed in could be either`a.b.test`, or `test`, or even `b.test`, `est`,`t`(if there is no conflict), etc.... 

Would be great if we can add examples like this in the comment to make this function easier to understand.

AliTajeldin:
For future, please don't create PR's on branches of branches.  Wait until the original branch is merged, then create the PR from master.  It is very hard to tell what actually changed in this branch without resorting to git command (and who really wants to do that 😄 )

laneb:
Yeah I'll take today's confusion as a lesson :stuck_out_tongue: next time I'll just put a comment in the ticket to say that the work is done but waiting on an upstream PR.

PR #1153, opened 2018-02-24
WIP: Spark 2.0 Branch

jacobdr:
**NOTE**
We need to create a Spark 2.0 branch from the current 2.1 tip, and that should be the real target of this PR

In retrospect I wish I would have fixed the 2.1, made a PR on that, and then branched from there for the backport....

History:
1. Branched off of 2.1
2. Merged current master
3. Dealt with conflicts, compilation errors, and API changes

### DIFF
--- a/admin/smv-release
+++ b/admin/smv-release
@@ -45,7 +45,7 @@ function parse_args()
 
   [ $# -ne 6 ] && echo "ERROR: invalid number of arguments" && usage 1
 
-  BUILD_DIR="$1"
+  BUILD_DIR=$(cd "$1"; pwd)
   SMV_BRANCH="$2"
   GITHUB_USER_TOKEN="$3"
   DOCKERHUB_USER_NAME="$4"
--- a/build.sbt
+++ b/build.sbt
@@ -2,13 +2,13 @@ name := "smv"
 
 organization := "org.tresamigos"
 
-version := "2.1-SNAPSHOT"
+version := "2.0-SNAPSHOT"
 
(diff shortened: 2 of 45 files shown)


jacobdr:
@AliTajeldin I forgot to change the `.spark_version` and `.smv_version` files, but want to discuss with you since we will need a Spark 2.0.0, even though 2.0 has been patched to 2.0.2

jacobdr:
In addition it looks like there are some real test failures from the Python side that I had failed to see when just running `sbt test`

Jeez, these Trais outputs are getting SUPER hard to read /cc @laneb

cloudysail:
I would suggest we merge master to 2.1 first(make sure everything runs well in 2.1), and then branch off of 2.1 to create 2.0. One thing at a time would make it a little bit easier.

I suppose you would later use `./admin/smv-release --new-branch ...` to do the 2.0 release. Early this month I did 2.2.0 release(although spark has latest 2.2.1, so I suppose your 2.0.0 should be fine). I fixed some issues for making a new branch release(rather than make a new release from an existing branch). One of the fixes is adding `function build_local_docker_image()` to `smv-release` so as to solve the spark version mismatch error when doing `itest`. You may need the same fix if you meet such issue. You can refer to my changes in 2.2 branch for reference.

AliTajeldin:
Closing this PR as we will probably not support 2.0 unless a client requests it.  We can redo 2.0 from 2.1 if needed in the future.

PR #1159, opened 2018-02-27
I1098 add validation results to metadata

### LINKED ISSUE #1098: Need a way to compare current validation results to historical ones
User wants to verify that, e.g., the number of failed records in this run does exceed the average for this dataset by more than some threshold proportion. For access to historical validation results, the obvious approach is to add the results to the default metadata and take advantage of the metadata history. We could also create a validation history in parallel to the metadata history, but it makes more sense to me to treat the validation result as another piece of metadata. If validation results are added to the metadata, we could just let the user compare results in `validateMetadata`. For a more tailored API we could also create a method to compare _just_ the validation results, but that could come later if needed.

laneb:
Resolves #1098.

### DIFF
--- a/src/main/python/smv/runinfo.py
+++ b/src/main/python/smv/runinfo.py
@@ -110,18 +110,38 @@ class SmvRunInfoCollector(object):
         return json.loads(java_result.toJson())['history']
 
     def show_report(self, ds_name=None, show_history=False):
+        """Print detailed report of information collected
+
+            Args:
+                ds_name (str): report only of named ds if not None
+                show_history (bool): include metadata history in report if True (default False)
+        """
         if ds_name is None:
             fqns = self.fqns()
         else:
             fqns = [self.jcollector.inferFqn(ds_name)]
         msg = 'datasets: %s' % fqns
+
+        def items_to_report(fqn):
+            validation = self.dqm_validation(fqn)
+            metadata = self.metadata(fqn)
+            # Remove validation results from metadata (if they exist) as we are reporting it above
+            try:
+                del metadata['_validation']
+            except:
+                pass
+
+            items = [("dqm validation", validation), ("metadata", metadata)]
+            if show_history:
+                history = self.metadata_history(fqn)
+                items.append(("metadata history", history))
+
+            return items
+
         for fqn_to_report in fqns:
             msg += '\n+ %s' % fqn_to_report
-            msg += '\n|- dqm validation:'
-            msg += '\n     ' + pformat(self.dqm_validation(fqn_to_report), indent=5)
-            msg += '\n|- metadata:'
-            msg += '\n     ' + pformat(self.metadata(fqn_to_report), indent=5)
-            if show_history:
-                msg += '\n|- metadata history:'
-                msg += '\n     ' + pformat(self.metadata_history(fqn_to_report), indent=5)
+            for name, value in items_to_report(fqn_to_report):
+                msg += '\n|- {}:'.format(name)
+                msg += '\n     ' + pformat(value, indent=5)
+
         print(msg)
--- a/src/main/scala/org/tresamigos/smv/SmvDataSet.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvDataSet.scala
@@ -18,7 +18,7 @@ import org.apache.spark.sql.{DataFrame, SaveMode}
 
 import org.apache.hadoop.fs.FileStatus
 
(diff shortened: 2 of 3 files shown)


src/main/scala/org/tresamigos/smv/SmvDataSet.scala
```diff
@@ -580,7 +586,7 @@ abstract class SmvDataSet extends FilenamePart {
     //Same as in persist, publish null string as a special value with assumption that it's not
     //a valid data value
     handler.saveAsCsvWithSchema(df, strNullValue = "_SmvStrNull_")
-    getOrCreateMetadata(Some(df)).saveToFile(app.sc, publishMetaPath(version))
+    getMetadata.saveToFile(app.sc, publishMetaPath(version))
```
cloudysail: Not sure if I missed anything. We used to pass `Some(df)` but now we call `getMetadata` in which we call `getOrCreateMetadata(None, None)`?

src/main/python/smv/runinfo.py
```diff
@@ -110,18 +110,38 @@ def metadata_history(self, ds_name):
...
         else:
             fqns = [self.jcollector.inferFqn(ds_name)]
         msg = 'datasets: %s' % fqns
+
+        def items_to_report(fqn):
+            validation = self.dqm_validation(fqn)
+            metadata = self.metadata(fqn)
+            # Remove validation results from metadata (if they exist) as we are reporting it above
+            try:
+                del metadata['_validation']
```
cloudysail: If we just do not add `("dqm validation", validation)` to items, but use `metadata['_validation']` to show the `DqmValiationResult`.  Would this work?

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
```diff
@@ -464,6 +467,9 @@ abstract class SmvDataSet extends FilenamePart {
     resMetadata.addDependencyMetadata(resolvedRequiresDS)
     dfOpt foreach {resMetadata.addSchemaMetadata}
     timestamp foreach {resMetadata.addTimestamp}
+    // note that validation result will be *excluded* from metadata provided to
```
AliTajeldin: Not sure I understand what this comment was intending to say.  Which validation result and when is the metadata validated?

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
laneb: The metadata has already been generated and saved to file by the time we publish, so `readPersistedMetadata` will return `Some` and we will not call `getOrCreateMetadata(None, None)`. This is very similar to the `getMetadataHistory` on the line below.  I reread this code a couple of times, and I don't this it's necessary to regenerate the metadata here. And in all likelihood, the user metadata will not be cached here because we didn't persist this output within this same transaction, so regenerating the metadata could be costly.

src/main/python/smv/runinfo.py
laneb: It would deduplicate the validation results in the report, but it would also make the report less readable. And the metadata is not the source of truth on validation, because when validation fails there will be validation results in `*.valid` but no metadata.

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
laneb: I'll add a comment to clarify that for both the metadata and metadata history.

src/main/scala/org/tresamigos/smv/SmvMetaData.scala
```diff
@@ -55,6 +57,19 @@ class SmvMetadata(val builder: MetadataBuilder = new MetadataBuilder) {
     builder.putString("_timestamp", dt.toString)
   }
 
+  /**
+   * Add validation result (including DQM state) to metadata
+   */
+   def addValidationResult(result: DqmValidationResult) = {
```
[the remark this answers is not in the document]
laneb: Sure.

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
laneb: `getOrCreateMetadata` is called to get metadata for metadata validation, and to get metadata to persist. At time of metadata validation, there is no DQM validation result to include in the metadata, as the validation hasn't completed. At the time the metadata is persisted, DQM validation has completed, so we have a result to include. We now have 3 possible degrees of 'completeness' for metadata: default metadata only _without_ validation result, user metadata + default metadata _without_ validation result, and user metadata + default metadata _with_ validation result. I'll put some clearer and better placed comments to explain this.

PR #1160, opened 2018-02-27
i1155 keep help shell function up to date automatically

### LINKED ISSUE #1155: Shell function help() should track new functions automatically
`smvshell.py`  includes a function `help` that prints a list of functions that are available in the shell, but these functions are rarely up to date and currently are very out of date. `help` should leverage `__all__` to identify the functions to document.

laneb:
Resolves #1155. In order to keep the URL to the API doc up to date, I needed access to the current SMV version, so I added a property `smvVersion` that reads from `.smv_version` to `SmvApp`. I also fixed the organization of the imports in the files I touched.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -28,6 +28,8 @@ parallelExecution in Test := false
 
 publishArtifact in Test := true
 
+envVars in Test := Map("SMV_HOME" -> ".")
+
 // Create itest task that runs integration tests
 val itest = TaskKey[Unit]("itest", "Run Integration Test")
 itest := {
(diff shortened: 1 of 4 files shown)


laneb:
Tests are failing because `SmvApp` uses the SMV_HOME env variable to figure out where `.smv_version` is, but SMV_HOME isn't set when running `sbt test`. Need to rethink.

cloudysail:
For tests, maybe we can use `System.getProperty("user.dir")` to get the SMV_HOME

laneb:
I think we can just make the `smvVersion` property an option which is `None` if there is no version.

AliTajeldin:
@laneb was this closed on purpose or by accident.

laneb:
Yeah, I thought I'd fixed the issue but there were still problems.

PR #1162, opened 2018-03-02
#1161 create API to get the latest history metadata

### LINKED ISSUE #1161: Support getting the latest history metadata
Right now we are using `smvserver.py->getMetadataJson()` which eventually calls `SmvDataSet.scala->getMetadata()` to construct information such as 'lastRunTime' of a module. This won't work if a module has never been run before or the output dir has been wiped out.

To solve this issue, we can utilize `SmvMetadataHistory->toJson()` to eventually construct our `lastRunTime`. The history metadata persisted in history folder has all the information we need, and the history would be empty if a module has never been run before.

We can even do optimization to avoid retrieving the whole history since the history could be big. We just need the latest history(the first item in the history array). To implement this, we can create a new function `SmvMetadataHistory->latestHistorytoJson()` to serialize the latest history metadata as JSON string, and expose it in `smvapp` and `smvserver`

cloudysail:
Fixes #1161

### DIFF
--- a/src/main/python/scripts/smvserver.py
+++ b/src/main/python/scripts/smvserver.py
@@ -185,9 +185,9 @@ def getMetadataJson(fqn):
     '''returns metadata given a fqn'''
     return SmvApp.getInstance().getMetadataJson("mod:{}".format(fqn))
 
-def getLatestMetadataHistoryJson(fqn):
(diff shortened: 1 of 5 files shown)


AliTajeldin:
LGTM

PR #1164, opened 2018-03-02
prepend library path as well 

### LINKED ISSUE #1163: Add library dir to py path as well
Spoke with @AliTajeldin this morning, and it has been decided that we want to standardize on adding `$PROJECT_DIR/library/` to the $PYTHON_PATH for SMV projects as well.

kmannislands:
Fixes #1163 while keeping it simple.

This is the cleanest way to resolve a bug in a downstream application that relies on discovering py modules in `<PROJECT_DIR>/library`


AliTajeldin:
Nice!

PR #1168, opened 2018-03-16
I1132 support user defined launch script

### LINKED ISSUE #1132: smv-run should accept driver script as argument
Use case: user wants to write a Python driver script that will decide contextually what module(s) to run with what configuration, etc. Needs SMV to create and configure SmvApp, then execute script.

Kai-Chen:
Closes #1132 

User script is supplied via the `--script` switch on the cmdline.

### DIFF
--- a/tools/smv-run
+++ b/tools/smv-run
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 #
 # Run one or more SMV modules / stages / app.
-# USAGE: smv-run [-h] [--script <script-path> [--script-opt <option>]*] <smv_app_args> [-- spark_args]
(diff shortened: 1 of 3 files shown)


tools/smv-run
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 #
 # Run one or more SMV modules / stages / app.
-# USAGE: smv-run [-h] smv_app_args -- spark_args
+# USAGE: smv-run [-h] [--script <script-path> [--script-opt <option>]*] <smv_app_args> [-- spark_args]
```
AliTajeldin: not using `script-opt` now.

PR #1182, opened 2018-04-10
i1169 quick run option

### LINKED ISSUE #1169: Option to skip metadata when running a module with smvApp.runModule
Sometimes I want to skip creating metadata e.g. when I run a module with `df` and just want a quick result. It would be convenient to be able to suppress the calculation of metadata when running a module.

laneb:
Resolves #1169. Will be handy for a couple of use cases at client site.


AliTajeldin:
nice!  need to resolve conflict to merge.

PR #1188, opened 2018-04-15
I1154 preserve helper docstrings

laneb:
@jacobdr 
screen shot 2018-04-15 at 1 46 30 am
Works with pydoc too.


ninjapapa:
need to `pip install decorator`.

AliTajeldin:
@laneb closed on purpose?  We had too many PR's closed by accident and lost track of them, so if you are closing on purpose please indicate so.

laneb:
Sorry, will do. I closed because the build was still failing even after adding `decorator` to the requirements`., Turns out we haven't configured travis to install our python dependencies before running tests.

PR #1197, opened 2018-04-17
Added SmvInputBase to SmvDataSet api

### LINKED ISSUE #1195: Make SmvInput python class more user friendly so user can easily define their own inputs
The use case is that other that the input connectors SMV is supporting, there might be all types of input sources which are not covered and can't be fully covered anyhow. The best way to address this could be providing an easier `SmvInput` ABC, so that user can easily extend for their own use cases.

As an example, we will use XML reader to test the design of the simpler `SmvInout` ABC.

This is a Type2 issue. Will design it within issue comments.

ninjapapa:
Closed #1195


src/main/python/smv/smvdataset.py
```diff
@@ -363,6 +366,46 @@ def run(self, df):
...
+        return None
+
+    def instanceValHash(self):
+        mTime = self.smvApp._jvm.SmvHDFS.modificationTime(self.fullpath())
+        pathHash = _smvhash(self.fullpath())
+        if (self.schema() is not None):
+            schemaHash = _smvhash(self.schema().simpleString())
+        else:
+            schemaHash = 0
+        return int(mTime + pathHash + schemaHash)
```
laneb: What happens if this result overflows a Java Int? I think in `sourceCodeHash` we use a bitmask to ensure that the result is in range.

ninjapapa:
Sounds like to fix `readAsDF` of `SmvInputWithScalaDS` from returning None is not an easy fix. Will close this PR to fix it.

src/main/python/smv/smvdataset.py
```diff
@@ -363,6 +366,46 @@ def run(self, df):
...
+class SmvInputFromFile(SmvInputBase):
+    """Base class for any input based on files on HDFS or local
+        Concrete class need to provide:
+            - fullpath (str): file full path with protocol
+            - readAsDF (DataFrame): file reading method
+            - schema (StructType): optional
+    """
+    @abc.abstractproperty
+    def fullpath(self):
+        """Full path to the input (file/dir or glob pattern)"""
```
AliTajeldin: need a `pass` at a minimum here.

src/main/python/smv/smvdataset.py
```diff
@@ -363,6 +366,46 @@ def run(self, df):
...
+
+    def schema(self):
+        """User specified schema
+            Returns:
+                (StructType): input data's schema
+        """
+        return None
+
+    def instanceValHash(self):
```
AliTajeldin: missing function doc.  Is this intended to compute hash of code, code plus upstream, etc?

src/main/python/smv/smvdataset.py
```diff
@@ -363,6 +366,46 @@ def run(self, df):
...
+        mTime = self.smvApp._jvm.SmvHDFS.modificationTime(self.fullpath())
+        pathHash = _smvhash(self.fullpath())
+        if (self.schema() is not None):
+            schemaHash = _smvhash(self.schema().simpleString())
+        else:
+            schemaHash = 0
+        return int(mTime + pathHash + schemaHash)
+
+
+class SmvInputWithScalaDS(SmvInputBase):
```
AliTajeldin: missing class doc.

src/main/python/smv/smvdataset.py
ninjapapa: Is it the case? All other abstractproperty do not have `pass`.

laneb: We don't have any linting set up but I think without `pass` this fails both `flake8` and `pylint`.

src/main/python/smv/smvdataset.py
laneb: No, `instanceValHash` (in contrast with `sourceCodeHash`) basically accounts for all the components of the `datasetHash` which are not code. `datasetHash` is only concerned about the target module and not its parents. `hashOfHash` is the one influenced by upstream modules.

src/main/python/smv/smvdataset.py
ninjapapa: OK. I will add `pass` here. Need to setup lint and update the best practice doc as a separate issue.

src/main/python/smv/smvdataset.py
```diff
@@ -363,20 +366,76 @@ def run(self, df):
...
+
+
+class SmvInputFromFile(SmvInputBase):
+    """Base class for any input based on files on HDFS or local
+        Concrete class need to provide:
+            - fullpath (str): file full path with protocol
+            - readAsDF (DataFrame): file reading method
+            - schema (StructType): optional
+    """
+    @abc.abstractproperty
```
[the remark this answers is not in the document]
ninjapapa: Sounds like we have been using `abstractproperty` in the wrong way for awhile. Let me create an issue for that, and also need to put some guidance in the best practice doc.

AliTajeldin:
@ninjapapa do you still need the branch or did you just forget to delete it?

PR #1211, opened 2018-04-20
I1207 use abstractmethod

### LINKED ISSUE #1207: Use abstractmethod instead of abstractproperty for interface methods
Here is how some code defined currently:

```python
class A(object):
    @abc.abstractproperty
    def path(self):
        pass
    def read(self):
        open(self.path(), "r")
        ......
```

If the concrete class of A is defined as the following:
```python
class B(A):
    def path(self):
        return "my_file"
```

Everything is fine. And it is actually how we documented our interface class A and method `path`. 

However since we claimed `path` is an `abstractproperty`, the concrete class of A should actually be defined as 
```python
class B(A):
    @property
    def path(self):
        return "my_file""
```

If the user really implements this way, we actually will run into errors like:
```
TypeError: 'str' object is not callable
```

because of the parenthesis after `path` in the `read` method.

## What we should do:
* only use method as the interface for our modules
* always use `abstractmethod` instead of `abstractproperty`

ninjapapa:
Fixed #1207 

Also formatting fix on smvdataset.py.

### DIFF
--- a/docs/dev/00_DevProcess/best_practice.md
+++ b/docs/dev/00_DevProcess/best_practice.md
@@ -41,6 +41,20 @@ When using Atom as editor, please install `linter-flake8` package. It will autom
 install dependencies. Also it need you to install flake8 on Python, so need to do
 `pip install flake8` in terminal.
 
+### Abstract decorators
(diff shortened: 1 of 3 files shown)


ninjapapa:
test error. close to fix.

PR #1217, opened 2018-04-21
I1212 path logic

### LINKED ISSUE #1212: Implement path logic in `SmvInputFromFile`
Child issue of #1201

ninjapapa:
Fixed #1212


ninjapapa:
Failed on py35, since using reduce. Close to fix.

PR #1232, opened 2018-04-28
#1230 create column helper function smvTimestampToStr

### LINKED ISSUE #1230: add column helper function to support smvTimestampToStr
Suppose the input is timestamp typed column(e.g. 1524848280), and we would like to output a String column which represents the local date/time/datetime of a given timezone and given format.

In spark there is a function called [pyspark.sql.functions.date_format](https://spark.apache.org/docs/latest/api/python/pyspark.sql.html#pyspark.sql.functions.date_format) doing the similar thing but it is very limited -- it always use the machine's system timezone.

For example:
```
class case2(smv.SmvCsvStringData):
    def schemaStr(self):
        return "ts:Timestamp[yyyyMMdd'T'HHmmssZ];tz:String"
    def dataStr(self):
        return '''
        20180428T025800+1000,UTC;
        20180428T025800+1000,America/Los_Angeles;
        20180428T025800+1000,Australia/Sydney
        '''
    def run(self, df):
        result = ( 
            df
            .withColumn("tsNumber",F.col("ts").cast("long"))
            .withColumn("SystemTZString",F.date_format(F.col("ts"),'YYYYMMdd HHmmssZ'))
            .withColumn("localdt",F.col('ts').smvTimestampToStr("UTC","yyyyMMdd:HHmmssz"))
            .withColumn("localdt2",F.col('ts').smvTimestampToStr(F.col("tz"),"yyyyMMdd:HHmmssz"))
        )
        return result
```
We would have output like this:
```
>>> df('test.case2').show(10,False)
+---------------------+-------------------+----------+--------------------+---------------------+---------------------+
|ts                   |tz                 |tsNumber  |SystemTZString      |localdt              |localdt2             |
+---------------------+-------------------+----------+--------------------+---------------------+---------------------+
|2018-04-27 09:58:00.0|UTC                |1524848280|20180427 095800-0700|20180427:165800+00:00|20180427:165800+00:00|
|2018-04-27 09:58:00.0|America/Los_Angeles|1524848280|20180427 095800-0700|20180427:165800+00:00|20180427:095800-07:00|
|2018-04-27 09:58:00.0|Australia/Sydney   |1524848280|20180427 095800-0700|20180427:165800+00:00|20180428:025800+10:00|
+---------------------+-------------------+----------+--------------------+---------------------+---------------------+

>>> df('test.case2').printSchema()
root
 |-- ts: timestamp (nullable = true)
 |-- tz: string (nullable = true)
 |-- tsNumber: long (nullable = true)
 |-- SystemTZString: string (nullable = true)
 |-- localdt: string (nullable = true)
 |-- localdt2: string (nullable = true)
```

cloudysail:
Fix #1230 
Internally, I create Joda DateTime Object with a long number(timestamp) and given timezone(string), and then use Joda's DateTimeFormatter to return the string with given format.

### DIFF
--- a/src/main/python/smv/helpers.py
+++ b/src/main/python/smv/helpers.py
@@ -1622,6 +1622,29 @@ class ColumnHelper(object):
         jc = self._jColumnHelper.smvStrToTimestamp(fmt)
         return Column(jc)
 
+    def smvTimestampToStr(self, timezone, fmt):
+        """Build a string from a timestamp and timezone
+
+            Args:
+                timezone (string or Column): the timezone follows the rules in https://www.joda.org/joda-time/apidocs/org/joda/time/DateTimeZone.html#forID-java.lang.String-
+                                             It can be a string like "America/Los_Angeles" or "+1000". If it is null, use current system time zone.
+                fmt (string): the format is the same as the Java `Date` format
+
+            Example:
+                >>> df.select(col("ts").smvTimestampToStr("America/Los_Angeles","yyyy-MM-dd HH:mm:ss"))
+
+            Returns:
+                (Column): StringType. The converted String with given format
+        """
+        if is_string(timezone):
+            jtimezone = timezone
+        elif isinstance(timezone, Column):
+            jtimezone = timezone._jc
+        else:
+            raise RuntimeError("timezone parameter must be either an string or a Column")
+        jc = self._jColumnHelper.smvTimestampToStr(jtimezone, fmt)
+        return Column(jc)
+
     def smvDay70(self):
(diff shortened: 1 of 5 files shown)


src/main/python/smv/helpers.py
```diff
@@ -1622,6 +1622,29 @@ def smvStrToTimestamp(self, fmt):
         jc = self._jColumnHelper.smvStrToTimestamp(fmt)
         return Column(jc)
 
+    def smvTimestampToStr(self, timezone, fmt):
+        """Build a string from a timestamp and timezone
```
AliTajeldin: The indent for the comment block seems to be off.

PR #1238, opened 2018-05-05
#1237 expose SmvDqmValidationError

### LINKED ISSUE #1237: Expose SmvDqmValidationError for end users to analysis
The use case is users want to do something by themselves(e.g send to report) when there is DQM error.

Right now if there is DQM error when we run our code, and if we use `try/catch` to check the Exception, it would be `Py4JJavaError` Object which has the error information as JSON string format. We could extract the information from the JSON String which tells us this Exception is really `SmvDqmValidationError` and also the detail `DQMValidationResult`. 

We don't want users to do all these extractions. We would like to handle that internally and returns a `SmvDqmValidationError` as dictionary type,  so that users can call it like this:
```
    try:
         smv_app = smv.SmvApp.getInstance()
         smv_app.runModuleByName("test")
    except SmvDqmValidationError as e:
        print(e.passed) # False
        print(e.dqmStateSnapshot["totalRecords"]) # 3
        print(e.dqmStateSnapshot["ruleErrors"]["c1 is null"]["total"]) # 1

class test(smv.SmvCsvStringData):
    def schemaStr(self):
        return "c1:String;c2:String"
    def dataStr(self):
        return '''
        a,x;
        ,y;
        c,z
        '''
    def run(self, df):
        return df
    def dqm(self):
        return (smv.dqm.SmvDQM()
                .add(smv.dqm.DQMRule(F.col("c1").isNotNull(),
                                     "c1 is null",
                                     smv.dqm.FailPercent(.005))
                 )
        )
```

cloudysail:
Fixes #1237

### DIFF
--- a/src/main/python/smv/error.py
+++ b/src/main/python/smv/error.py
@@ -13,7 +13,18 @@
 """
 Errors thrown by SMV
 """
+import json
 
 class SmvRuntimeError(RuntimeError):
     def __init__(self,msg):
         super(SmvRuntimeError,self).__init__(msg)
+
+class SmvDqmValidationError(SmvRuntimeError):
+    """ This class takes the DqmValidationResult(JSON String) msg, transfers it to a dict,
+        and then set self attributes to contains all the information from the dict.
+    """
+    def __init__(self,msg):
+        super(SmvRuntimeError,self).__init__(msg)
+        error_dict = json.loads(msg)
+        for key in error_dict:
+            setattr(self, key, error_dict[key])
--- a/src/main/python/smv/error.py
+++ b/src/main/python/smv/error.py
@@ -20,11 +20,8 @@ class SmvRuntimeError(RuntimeError):
         super(SmvRuntimeError,self).__init__(msg)
 
 class SmvDqmValidationError(SmvRuntimeError):
-    """ This class takes the DqmValidationResult(JSON String) msg, transfers it to a dict,
-        and then set self attributes to contains all the information from the dict.
+    """ This class has an instance of dqmValidationResult(dict)
     """
-    def __init__(self,msg):
-        super(SmvRuntimeError,self).__init__(msg)
-        error_dict = json.loads(msg)
-        for key in error_dict:
-            setattr(self, key, error_dict[key])
+    def __init__(self, dqmValidationResult):
+        super(SmvDqmValidationError,self).__init__(json.dumps(dqmValidationResult))
+        self.dqmValidationResult = dqmValidationResult
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -144,14 +144,15 @@ class SmvApp(object):
 
     def exception_handling(func):
         """ Decorator function to catch Exception and raise SmvDqmValidationError if any.
-            Otherwise just pass the original exception
+            Otherwise just pass through the original exception
         """
         def func_wrapper(*args, **kwargs):
             try:
                 return func(*args, **kwargs)
             except Exception as e:
-                if(e.java_exception and e.java_exception.getClass().getName() == "org.tresamigos.smv.SmvDqmValidationError"):
-                    raise SmvDqmValidationError(e.java_exception.getMessage())
+                if (e.java_exception and e.java_exception.getClass().getName() == "org.tresamigos.smv.dqm.SmvDqmValidationError"):
+                    dqmValidationResult = json.loads(e.java_exception.getMessage())
+                    raise SmvDqmValidationError(dqmValidationResult)
                 else:
                     raise e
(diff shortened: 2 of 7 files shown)


src/main/python/smv/smvapp.py
```diff
@@ -141,6 +141,21 @@ def __init__(self, arglist, _sc = None, _sqlContext = None):
         # Initialize DataFrame and Column with helper methods
         smv.helpers.init_helpers()
 
+
+    def exception_handling(func):
```
AliTajeldin: nice!

src/main/python/smv/error.py
```diff
@@ -13,7 +13,18 @@
...
         super(SmvRuntimeError,self).__init__(msg)
+
+class SmvDqmValidationError(SmvRuntimeError):
+    """ This class takes the DqmValidationResult(JSON String) msg, transfers it to a dict,
+        and then set self attributes to contains all the information from the dict.
+    """
+    def __init__(self,msg):
+        super(SmvRuntimeError,self).__init__(msg)
+        error_dict = json.loads(msg)
+        for key in error_dict:
```
AliTajeldin: Had a discussion with @jacobdr about this.  I don't think we should spread the contents of the data validator json representation into the exception object directly.  Once we start moving more of the Scala code into python, we would want to use this python class to raise DQM validation errors from python directly.  This will make it difficult to create such an exception.  Perhaps we should mimic the Scala side fully.  That is, the exception will have a ValidationResult class instance (or whatever it is called on the scala side).

src/main/python/smv/smvapp.py
```diff
@@ -141,6 +141,21 @@ def __init__(self, arglist, _sc = None, _sqlContext = None):
...
+        """ Decorator function to catch Exception and raise SmvDqmValidationError if any.
+            Otherwise just pass the original exception
+        """
+        def func_wrapper(*args, **kwargs):
+            try:
+                return func(*args, **kwargs)
+            except Exception as e:
+                if(e.java_exception and e.java_exception.getClass().getName() == "org.tresamigos.smv.SmvDqmValidationError"):
```
AliTajeldin: space after `if`

src/main/python/smv/smvapp.py
```diff
@@ -141,6 +141,21 @@ def __init__(self, arglist, _sc = None, _sqlContext = None):
...
+                    raise SmvDqmValidationError(e.java_exception.getMessage())
```
AliTajeldin: I don't see a way to ensure that the `message` is a valid json as we have discussed.

cloudysail:
@AliTajeldin Code fixed according to your feedback. Ready for review.

src/main/python/smv/error.py
```diff
@@ -20,11 +20,8 @@ def __init__(self,msg):
         super(SmvRuntimeError,self).__init__(msg)
 
 class SmvDqmValidationError(SmvRuntimeError):
-    """ This class takes the DqmValidationResult(JSON String) msg, transfers it to a dict,
-        and then set self attributes to contains all the information from the dict.
+    """ This class has an instance of dqmValidationResult(dict)
```
AliTajeldin: how do I raise this exception in python?

cloudysail: Right now I'm using 
```
dqmValidationResult = json.loads(e.java_exception.getMessage())
raise SmvDqmValidationError(dqmValidationResult)
```
The `dqmValidationResult` is a json structure which comes from the scala side. Later if we re-create the json struture in python side, then we can pass `dqmValidationResult` from python directly to raise `SmvDqmValidationError`.

cloudysail: Created #1239 to create `DqmValidationResult` and all related classes in python

src/main/python/smv/smvapp.py
```diff
@@ -141,6 +141,22 @@ def __init__(self, arglist, _sc = None, _sqlContext = None):
...
+
+    def exception_handling(func):
+        """ Decorator function to catch Exception and raise SmvDqmValidationError if any.
+            Otherwise just pass through the original exception
+        """
+        def func_wrapper(*args, **kwargs):
+            try:
+                return func(*args, **kwargs)
+            except Exception as e:
+                if (e.java_exception and e.java_exception.getClass().getName() == "org.tresamigos.smv.dqm.SmvDqmValidationError"):
```
laneb: Pretty sure this will cause an `AttributeError` if the exception in `func` is occurring on this side of the py4j Gateway. I think it would specifically catch py4j exceptions above.

cloudysail: @laneb you're totally right! Thanks for catching this! I'll fix it in #1240

PR #1245, opened 2018-05-16
add option to python openCsv function to validate CSV file

### LINKED ISSUE #1244: Add option to Python openCsv shell function to validate CSV file
2 issues:
1. Python `openCsv` doesn't have this option
2. Scala `openCsv` has this option but it doesn't work

laneb:
resolves #1244

### DIFF
--- a/src/main/scala/org/tresamigos/smv/shell/ShellCmd.scala
+++ b/src/main/scala/org/tresamigos/smv/shell/ShellCmd.scala
@@ -163,21 +163,8 @@ object ShellCmd {
     SmvApp.app.publishModuleToHiveByName(dsName, runConfig, collector)
   }
 
-  /**
-   * Read in a Csv file as DF
-  **/
-  def openCsv(path: String, ca: CsvAttributes, parserCheck: Boolean,
-    collector: SmvRunInfoCollector=new SmvRunInfoCollector): DataFrame = {
-
-    /** isFullPath = true to avoid prepending data_dir */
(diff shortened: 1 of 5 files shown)


src/main/scala/org/tresamigos/smv/shell/ShellCmd.scala
```diff
@@ -163,21 +163,8 @@ object ShellCmd {
...
-    /** isFullPath = true to avoid prepending data_dir */
-    object file extends SmvCsvFile(path, ca, null, true) {
-      override val forceParserCheck   = false
-      override val failAtParsingError = parserCheck
-    }
-    file.rdd(collector=collector)
-  }
-
-  def openCsv(path: String): DataFrame = openCsv(path, null, false)
+  def openCsv(path: String, ca: CsvAttributes, validate: Boolean): DataFrame =
```
cloudysail: We do want to remove parameter `collector: SmvRunInfoCollector=new SmvRunInfoCollector` from this API?

laneb: Yeah, I don't think there's any use for having the `collector` as an argument in a shell function. And the argument was added after the Scala shell was largely abandoned, so no one's ever used it. If anyone ever needs it, it's still available and the `SmvApp` method.

cloudysail: Sounds good.

PR #1248, opened 2018-05-23
add user_libraries prop to conf, decache py mods 

### LINKED ISSUE #1163: Add library dir to py path as well

kmannislands:
Fixes #1163
- Adds a potential prop to Smv config for 'user_libraries'
- Uses that prop with existing 'decache python module' strategy in `DataSetRepo` so that UDL's are reloaded.


ninjapapa:
Really nice!

PR #1251, opened 2018-05-25
I1231 fix metadata causes dqm double count

laneb:
The key was to generate the metadata based on the df _before_ attaching DQM. 

@jacobdr fyi


laneb:
A developer at a client site has just pointed out to me that this change will metadata to be based on the module output _before_ fixing/dropping records, which isn't the behavior you'd probably expect. Will need to have a bit more of a think on this. Instead of trying to force the metadata to occur before DQM validation, it may be more practical to accept relying on arithmetic to infer the time spent on DQM validation alone (subtract metadata time from total elapsed DQM time).

PR #1253, opened 2018-05-25
I1252 undefined runconfig return none

skrauselevy:
Now ```smvGetRunConfig, smvGetRunConfigAsInt, smvGetRunConfigAsBool``` return None if the runConfig does not exist.

### DIFF
--- a/src/test/python/testSmvFramework/stage/modules.py
+++ b/src/test/python/testSmvFramework/stage/modules.py
@@ -47,38 +47,39 @@ class D4(SmvCsvStringData, SmvRunConfig):
     def schemaStr(self):
         return "a:String;b:Integer"
     def dataStr(self):
-        if(self.smvGetRunConfig('s') == "s1"):
-            l1 = "test1_s1,1"
+        testVals = [];
+        if (self.smvGetRunConfig('s') == "s1"):
+            testVals.append("test1_s1,1")
         else:
-            l1 = "test1_not_s1,2"
+            testVals.append("test1_not_s1,2")
 
-        if(self.smvGetRunConfigAsInt('i') == 2):
-            l2 = "test2_i2,3"
+        if (self.smvGetRunConfigAsInt('i') == 2):
+            testVals.append("test2_i2,3")
         else:
-            l2 = "test2_not_i2,4"
+            testVals.append("test2_not_i2,4")
 
-        if(self.smvGetRunConfigAsBool('b')):
-            l3 = "test3_b,5"
+        if (self.smvGetRunConfigAsBool('b')):
+            testVals.append("test3_b,5")
         else:
-            l3 = "test3_not_b,6"
+            testVals.append("test3_not_b,6")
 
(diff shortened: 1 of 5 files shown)


src/test/python/testSmvFramework/stage/modules.py
```diff
@@ -62,7 +62,23 @@ def dataStr(self):
         else:
             l3 = "test3_not_b,6"
 
-        return l1 + ";" + l2 + ";" + l3
+        # Tests to make sure that if the runconfig doesn't exist, None is returned
+        if(self.smvGetRunConfig('c') is None):
```
laneb: You can skip the parentheses. Makes it a little more pythonic / nicer to read.

src/test/python/testSmvFramework/stage/modules.py
```diff
@@ -62,7 +62,23 @@ def dataStr(self):
...
+            l5 = "test5_undefined_one,9"
+        else:
+            l5 = "test5_defined_one,10"
+
+        if(self.smvGetRunConfigAsBool('bool') is None):
+            l6 = "test6_undefined_bool,11"
+        else:
+            l6 = "test6_defined_bool,12"
+
+        return l1 + ";" + l2 + ";" + l3 + ";" + l4 + ";" + l5 + ";" + l6
```
laneb: if you collect the vals in a list 
```
vals = []
...
if(self.smvGetRunConfig('c') is None):
    vals.append("test4_undefined_c,7")
```
then you can just join them at the end
```
";".join(vals)
```

src/test/scala/org/tresamigos/smv/SmvAppTest.scala
```diff
@@ -57,6 +57,9 @@ package org.tresamigos.smv {
 
       app.runModule(C.urn, runConfig = Map("sample" -> "3pct"))
       assert(app.smvConfig.getRunConfig("sample") === "3pct")
+
+      // test to make sure undefined runConfig returns null not an exception
+      assert(app.smvConfig.getRunConfig("sampleTwo") === null)
```
laneb: Shouldn't this raise an error since the key `sampleTwo` isn't declared in the config?

From the issues description:
```
In the case of running a module that is trying to access a key|value pair that does not exist the below error is thrown
```

src/test/python/testSmvFramework/stage/modules.py
laneb: You can skip the parentheses. Makes it a little more pythonic / nicer to read.

src/test/scala/org/tresamigos/smv/SmvAppTest.scala
skrauselevy: No, It is supposed to return None. That is the current behavior, the wanted behavior is return None.

laneb: Sorry, misread the ticket.

PR #1257, opened 2018-05-30
Use Libs in Hash

### LINKED ISSUE #1247: Incorporate user libraries into SmvDataSet hash
#1163 adds `user_libraries` to the app props and supports decaching the python modules.
This ticket is to take that in to account in the SmvDataSet hash to know to rerun modules when a library that they depend on has changed.

kmannislands:
Fixes #1247

Factor external code's source into DataSet hash.

This is intended to be used with project user-defined libraries but can technically be used on almost any python object (anything that `inspect.getsource()` can be called on).

### DIFF
--- a/docs/user/app_config.md
+++ b/docs/user/app_config.md
@@ -119,7 +119,7 @@ Note that for sequence/list type parameters (e.g. smv.stages), a "," or ":" can 
 <td>smv.user_libraries</td>
 <td>empty</td>
 <td>Optional</td>
-<td>A list of external (not SMV DataSet subclass instances) modules, functions and/or objects that should be reloaded with the project</td>
+<td>A list of external python modules, functions and/or objects that should be reloaded with the project</td>
 </tr>
 
 <tr>
--- a/docs/user/app_config.md
+++ b/docs/user/app_config.md
@@ -116,6 +116,13 @@ Note that for sequence/list type parameters (e.g. smv.stages), a "," or ":" can 
 </tr>
 
 <tr>
+<td>smv.user_libraries</td>
+<td>empty</td>
+<td>Optional</td>
+<td>A list of external (not SMV DataSet subclass instances) modules, functions and/or objects that should be reloaded with the project</td>
+</tr>
+
+<tr>
 <th colspan="4">Data Directories Parameters</th>
 </tr>
 
--- a/docs/user/smv_dataset.md
+++ b/docs/user/smv_dataset.md
@@ -40,7 +40,7 @@ This means cached data will be invalidated if the external code is changed, etc.
 
 There are two steps to setting this up for your project:
 
-1. Add the library to your app conf via either of the files under `conf/` or the command line under the `smv.user_libraries` property.
+1. Add the `smv.user_libraries` prop to your [app configuration](./app_config.md).
 
 2. Import the library in all files that depend on it (like `import udl as lib`) and then in the class defs of DataSets that depend on it, implement the `requiresLib()` method and make it return an array of the libraries that that DataSet depends on (for the first example, `requiresLib()` would return `[lib]`).
 
--- a/docs/user/smv_dataset.md
+++ b/docs/user/smv_dataset.md
@@ -36,9 +36,22 @@ Any of the classes listed above can implement a `requiresLib()` function to defi
 
 This means cached data will be invalidated if the external code is changed, etc.
 
+> **Limitations:** For python modules and packages, the `requiresLib()` method is limited to registering changes on the main file of the package (for module `foo`, that's `foo.py`, for package `bar`, that's `bar/__init__.py`). This means that if a module or package imports other modules, the imported module's changes will not impact DataSet hashes.
+
+There are two steps to setting this up for your project:
+
+1. Add the library to your app conf via either of the files under `conf/` or the command line under the `smv.user_libraries` property.
+
+2. Import the library in all files that depend on it (like `import udl as lib`) and then in the class defs of DataSets that depend on it, implement the `requiresLib()` method and make it return an array of the libraries that that DataSet depends on (for the first example, `requiresLib()` would return `[lib]`).
+
 **Python**
 
 ```py
+# in smv-app-conf.props:
+smv.stages = stage
+smv.user_libraries = lib, pandas
+
+# in src/main/python/stage/someFile.py:
 import some_library as lib
 import pandas
 
--- a/docs/user/smv_dataset.md
+++ b/docs/user/smv_dataset.md
@@ -20,15 +20,42 @@ directly, instead, they will use all the detailed types of `SmvDataSet`.
 
 In this document we will cover the following basic `SmvDataSet`s,
 
- | Scala | Python
---- | --- | ---
-**Input** | `SmvCsvFile` | `SmvCsvFile`
-          | `SmvMultiCsvFiles` | `SmvMultiCsvFiles`
-          | `SmvCsvStringData` | `SmvCsvStringData`
-          | `SmvHiveTable` | `SmvHiveTable`
-**Intermediate** | `SmvModule` | `SmvModule`
-**Output** (mix in) | `SmvOutput` | `SmvOutput`
+|                     | Scala              | Python             |
+|---------------------|--------------------|--------------------|
+| **Input**           | `SmvCsvFile`       | `SmvCsvFile`       |
+|                     | `SmvMultiCsvFiles` | `SmvMultiCsvFiles` |
+|                     | `SmvCsvStringData` | `SmvCsvStringData` |
+|                     | `SmvHiveTable`     | `SmvHiveTable`     |
+| **Intermediate**    | `SmvModule`        | `SmvModule`        |
+| **Output (mix in)** | `SmvOutput`        | `SmvOutput`        |
+|                     |                    |                    |
 
+### External Code Dependency (python only)
+
+Any of the classes listed above can implement a `requiresLib()` function to define code that isn't SMV DataSets (like an external or user-defined library) that should be considered like it is part of the source of the DataSet itself.
+
+This means cached data will be invalidated if the external code is changed, etc.
+
+**Python**
+
+```py
+import some_library as lib
+import pandas
+
+class MyModule(smv.SmvModule):
+    """mod description"""
+    # define dependency on other SmvDataset(s)
+    def requiresDS(self):
+        return [Mod1, Mod2]
+
+    # define dependency on arbitary external code
+    # ie: libraries defined in your project or from pip
+    def requiresLib(self):
+        return [lib, pandas]
+
+    def run(self, i):
+        ...
+```
 
 ## Input SmvDataSet
 
--- a/src/main/python/smv/smvdataset.py
+++ b/src/main/python/smv/smvdataset.py
@@ -48,6 +48,14 @@ def _stripComments(code):
     code = str(code)
     return re.sub(r'(?m)^ *(#.*\n?|[ \t]*\n)', '', code)
 
+def _sourceHash(module):
+    src = inspect.getsource(module)
+    src_no_comm = _stripComments(src)
+    # DO NOT use the compiled byte code for the hash computation as
+    # it doesn't change when constant values are changed.  For example,
+    # "a = 5" and "a = 6" compile to same byte code.
+    # co_code = compile(src, inspect.getsourcefile(cls), 'exec').co_code
+    return _smvhash(src_no_comm)
 
 class SmvOutput(object):
     """Mixin which marks an SmvModule as one of the output of its stage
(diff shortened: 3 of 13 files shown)


kmannislands:
@jacobdr, documentation connoisseur--is there a better place I should doc this?

src/test/python/testModuleHash.py
```diff
@@ -35,13 +35,19 @@ def __init__(self, smvApp, path, fqn):
...
             self.path = path
             self.fqn = fqn
+        
+        def lib_path(self):
+            # use a realistic library dir, which we'll also add to py path
+            return self.path + "/library"
 
         def __enter__(self):
             sys.path.insert(1,self.path)
+            sys.path.insert(1,self.lib_path())
```
laneb: We should set the app dir instead of manipulating the `sys.path` directly. Then it's slightly more representative of a real use case, and there's no need to touch the `lib_path`.

laneb:
I think we need a couple of additional tests to guarantee the correctness of the hash behavior: 

1. Test that the hash _doesn't_ change if the library doesn't change
2. Test the changed and unchanged cases for a function in place of a library

The source hash has historically had a lot of bugs, so it'll be valuable to have fairly extensive coverage here.

src/main/python/smv/smvdataset.py
```diff
@@ -98,6 +106,17 @@ def requiresDS(self):
             Returns:
                 (list(SmvDataSet)): a list of dependencies
         """
+    
+    def requiresLib(self):
+        """User-specified list of user-defined (not external) library dependencies
```
laneb: Docstring conflicts with smv_dataset.md - should the user include external libraries or not? Makes sense to me that  the hash should change if the external libraries change.

src/main/python/smv/smvdataset.py
```diff
@@ -98,6 +106,17 @@ def requiresDS(self):
...
+
+            Override this method to assist in re-running this module based on changes
+            in other python files.
+
+            Returns:
+                (list(Module)): a list of library dependencies
```
laneb: Module should be lower-case to match the name of the type.

src/test/python/testModuleHash.py
```diff
@@ -20,7 +20,7 @@
 class ModuleHashTest(SmvBaseTest):
     @classmethod
     def smvAppInitArgs(cls):
-        return ["--smv-props", "smv.stages=stage"]
+        return ["--smv-props", "smv.stages=stage", "smv.user_libraries=udl:same"]
```
ninjapapa: Is the `smv. user_libraries` parameter described in any user doc?

ninjapapa: Is `smv. user_libraries` covered in any user doc?

ninjapapa:
Does `inspect.getsource` on `panads` really work?

src/test/python/testModuleHash.py
kmannislands: @ninjapapa no, we were initially planning on leaving it undocumented since we anticipated it being primarily used programmatically by downstream app. 
We have since realized that the feature is useful and simple enough to consume that we should doc. Will add some doc on the `user_libraries` prop as well.

ninjapapa: So basically if I have module requiresLib `pandas` and `myownlib`, I need to specify them in my `smv-app-conf` file. Is it the case? If so we do have to document it in some user doc.

laneb:
@ninjapapa Yes:
```
>>> import pandas
>>> import inspect
>>> inspect.getsource(pandas)
'# pylint: disable-msg=W0614,W0401,W0611,W0622\n\n# flake8: noqa\n\n__docformat__ = \'restructuredtext\'\n\n# Let users know if they\'re missing any of our hard ..
```
(ellipsis is mine)

but since you bring it up, it occurs to me that the usefulness of depending on the source of an external library like `pandas` is questionable, mostly because `inspect.getsource` will return only a single file's contents, either the module itself if it is a module or the `__init__.py` if it is a  package.

laneb:
We should at least document the limitations - we'll face the same limitations for user libraries.

kmannislands:
@laneb added some doc on limitations in the last commit.
LMK if my python terminology is accurate

docs/user/app_config.md
```diff
@@ -115,6 +115,13 @@ Note that for sequence/list type parameters (e.g. smv.stages), a "," or ":" can
 <td>A list of runtime configuration keys user can specify, and accessed through the `smvGetRunConfi` method of `SmvRunConfig` mix-in</td>
 </tr>
 
+<tr>
+<td>smv.user_libraries</td>
+<td>empty</td>
+<td>Optional</td>
+<td>A list of external (not SMV DataSet subclass instances) modules, functions and/or objects that should be reloaded with the project</td>
```
laneb: Would be clearer to clarify what a module is in this context (a python file) instead of what it's not. Also, I think user_libraries is just supposed to indicate the location of user library right? I.e. this would never include external libraries or functions/objects. I think maybe the documentation is getting a bit mixed up.

docs/user/smv_dataset.md
```diff
@@ -36,9 +36,22 @@ Any of the classes listed above can implement a `requiresLib()` function to defi
...
+1. Add the library to your app conf via either of the files under `conf/` or the command line under the `smv.user_libraries` property.
+
+2. Import the library in all files that depend on it (like `import udl as lib`) and then in the class defs of DataSets that depend on it, implement the `requiresLib()` method and make it return an array of the libraries that that DataSet depends on (for the first example, `requiresLib()` would return `[lib]`).
+
 **Python**
 
 ```py
+# in smv-app-conf.props:
+smv.stages = stage
+smv.user_libraries = lib, pandas
```
laneb: Not pandas, right?

docs/user/smv_dataset.md
```diff
@@ -36,9 +36,22 @@ Any of the classes listed above can implement a `requiresLib()` function to defi
 
 This means cached data will be invalidated if the external code is changed, etc.
 
+> **Limitations:** For python modules and packages, the `requiresLib()` method is limited to registering changes on the main file of the package (for module `foo`, that's `foo.py`, for package `bar`, that's `bar/__init__.py`). This means that if a module or package imports other modules, the imported module's changes will not impact DataSet hashes.
+
+There are two steps to setting this up for your project:
+
+1. Add the library to your app conf via either of the files under `conf/` or the command line under the `smv.user_libraries` property.
```
laneb: Instead of explaining how to add properties to the conf just link the doc on the app conf

docs/user/smv_dataset.md
```diff
@@ -36,9 +36,22 @@ Any of the classes listed above can implement a `requiresLib()` function to defi
 
 This means cached data will be invalidated if the external code is changed, etc.
 
+> **Limitations:** For python modules and packages, the `requiresLib()` method is limited to registering changes on the main file of the package (for module `foo`, that's `foo.py`, for package `bar`, that's `bar/__init__.py`). This means that if a module or package imports other modules, the imported module's changes will not impact DataSet hashes.
```
laneb: This is good - let's also add it to the docstring for `requiresLib`.

src/test/python/testModuleHash.py
laneb: There would be no need to add an external library like `pandas` to `smv.user_libraries`, right? Doesn't that just list the location of the code for UDLs?

ninjapapa:
LGTM as long as comments from @laneb get addressed.

docs/user/smv_dataset.md
kmannislands: This prop is used in the DataSet repo to ensure that the modules enumerated in `smv.user_libraries` are removed from python's cache, so I think external libs would have to be listed too.
Agreed that this prop has expanded to be more than `user_libraries`, trying to think of another name... Just `libraries`? `external_py_modules`?

kmannislands: Could probably also get rid of the property altogether and instead call the `requiredLib()` method on modules as we de-cache them.

laneb: Was not aware we intended for external libraries to be reloaded...what's a scenario for that? Long lived application wants to leverage updated pandas version?

laneb: I think there a would be a significant risk of mismatched external lib versions. Not sure what the consequences could be.

kmannislands:
I have two approaches here:
1. Rename the `smv.user_libraries` conf prop to make it more general/accurate

2. I have an approach where there is no need to enumerate these libraries/external modules in the app conf at all. This is achieved by inspecting modules as we de-cache them in `datasetrepo.py`, getting classes defined in them, and checking the result of calling each one's `requiresLib` function.

The first approach is more performant in brief testing, but the second approach requires less boilerplate and could prevent tricky bugs where a library is listed in a `requiresLib` function, but not in the app props.

ninjapapa:
I surly prefer the 2nd approach.

kmannislands:
@ninjapapa I'll open a second PR with that approach shortly.

kmannislands:
@ninjapapa, @laneb opened #1263 for the other approach. Looking forward to feedback.

kmannislands:
Closing this PR in favor of approach number 3.

kmannislands:
Reopening, turns out simple _was_ best.

kmannislands:
Only consideration left here for me is that the app conf prop `user_libraries` could have a more technically accurate name.
If anyone support a different name like:
- `additional_modules`: clarifies that they are python modules, additional to the smv app codebase that should be considered as part of the code
- `libraries`: nice because the expected dir is `library` so its internally consistent. Also describes the superset of user libs and external libs accurately
Then I will rename it. Otherwise, should be good to go.

laneb:
Opinion: messages like
> fix error introduced in rebase #1247
defeats the whole purpose of rebasing. If you're going to rebase, you should re-rebase if you make mistakes.

laneb:
Has the code in this PR changed besides the rebase+fix?

laneb:
Can you add a dev doc describing the approaches we discussed and why we settled on this one? Will be helpful down the road.

kmannislands:
@laneb good call will do

kmannislands:
@laneb doc'd

ninjapapa:
To confirm my understanding: Let's say I have a project with a 'libs' folder with a '__init__.py' file, to make sure my module's hash change when libs code change, I need to 
- add 'libs' to smv-user-conf
- add 'requires_libs' to the module
- define some 'version' number in the '__init__.py' file
- change that version number whenever I made some code change in the libs package

kmannislands:
@ninjapapa yes, that would work.
Hypothetically, you had a tool of some sort generating the module and library code for project with structure like:
```
src
└── main
    └── python
        └── employment
            ├── __init__.py
            └── employment.py
library
└── user_lib
    ├── __init__.py
    └── module.py
```

Then instead of adding:
```py
def requiresLib(self):
  return [user_lib]
```
We could design the tool to generate:
```py
def requiresLib(self):
  return [user_lib, user_lib.module]
```
Then, as long as `smv.app-conf.props` has:
```py
# ...other app conf
smv.user_libraries = user_lib
```
The library should affect the module hash and be reloaded for transactions _without the need to increment a version number_ in `__init__.py`.

Additionally, this hypothetical tool could simply generate flattened `library` structures in the first place.

laneb:
We _could_ make the hash a bit smarter about taking child packages into consideration. Iterate over the subpackages of the required lib and hash their source code as well. Let's create a separate ticket to consider that. Doesn't need to block this PR

AliTajeldin:
"Hypothetically".  LOL 😄

kmannislands:
@AliTajeldin @laneb is this good to merge then?

AliTajeldin:
LGTM

docs/user/smv_dataset.md
```diff
@@ -20,15 +20,55 @@ directly, instead, they will use all the detailed types of `SmvDataSet`.
...
+class MyModule(smv.SmvModule):
+    """mod description"""
+    # define dependency on other SmvDataset(s)
+    def requiresDS(self):
+        return [Mod1, Mod2]
+
+    # define dependency on arbitary external code
+    # ie: libraries defined in your project or from pip
+    def requiresLib(self):
+        return [lib, pandas]
```
cloudysail: Maybe we could define two Modules here, so that show the requiresLib() could be a subset of all imports. I think that's why we really need `requiresLib()`
```
import some_library1 as lib1
import some_library2 as lib2
import pandas
class MyModule1(smv.SmvModule):
  def requiresLib(self):
    return [lib1, pandas]
class MyModule2(smv.SmvModule):
  def requiresLib(self):
    return [lib2]
```

kmannislands: @cloudysail that's an excellent idea. We were motivated in the design by the need to support old code bases where some files shouldn't be reloaded, but it is sorta just implied in our doc.
I'll create a ticket to specify explicitly what to do to avoid having py modules reloaded (or reload a subset of py modules)

PR #1260, opened 2018-06-01
I1231 metadata causes dqm double count

### LINKED ISSUE #1231: User metadata causes DQM rules failures to be double counted
Cause is that we calculate metadata after applying the DQM tasks to the output `DataFrame` but before actually validating `DataFrame`.

laneb:
Fixes #1231.


ninjapapa:
Very nice.

src/main/scala/org/tresamigos/smv/SmvMetaData.scala
```diff
@@ -197,30 +197,3 @@ object SmvMetadataHistory {
     new SmvMetadataHistory(Array.empty)
   }
 }
-
-/**
- * Policy for validating a module's current metadata against its historical
- * metadata. This policy is added to every module's DQM
- */
-class DQMMetadataPolicy(ds: SmvDataSet) extends dqm.DQMPolicy{
```
cloudysail: Will need to update dev doc since we removed this class. If not now, maybe after we rebuild the DQM

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
```diff
@@ -551,19 +545,22 @@ abstract class SmvDataSet extends FilenamePart {
...
-      getOrCreateUserMetadata(df)
       val (validationResult, validationDuration) =
         doAction(f"VALIDATE DATA QUALITY") {dqmValidator.validate(df, hasAction, moduleValidPath())}
       dqmTimeElapsed = Some(validationDuration)
 
       val metadata = getOrCreateMetadata(Some(df), Some(validationResult))
       // must read metadata from file (if it exists) before deleting outputs
       val metadataHistory = getMetadataHistory
+
+      val validationRes: Option[String] = validateMetadata(metadata, metadataHistory.historyList)
```
cloudysail: Shall we have a comment like `Metadata to be validated includes user metadata and  DqmValidationResult`

src/main/scala/org/tresamigos/smv/dqm/SmvDQM.scala
```diff
@@ -228,7 +228,7 @@ class DQMValidator(dqm: SmvDQM, persistable: Boolean) {
       toConsole(res)
 
     // persist if result is not empty or forced an action
```
cloudysail: Need to update the old comment

src/test/python/testSmvFramework.py
```diff
@@ -186,14 +190,8 @@ def test_metadata_includes_user_metadata(self):
 
     def test_metadata_validation_failure_causes_error(self):
         fqn = "metadata_stage.modules.ModWithFailingValidation"
-        try:
+        with self.assertRaisesRegexp(Py4JJavaError, "SmvMetadataValidationError"):
```
cloudysail: Shall we also create `SmvDqmValidationError` in python like `SmvDqmValidationError ` so that users can catch it rather than handling `Py4JJavaError ` directly?

laneb: I can do that, although I don't think we have a use case for it. Is the intent to follow this pattern for all `SmvRuntimExceptions` to be intercepted in this way?

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
laneb: Done.

src/main/scala/org/tresamigos/smv/SmvMetaData.scala
laneb: The dev doc doesn't discuss `DQMMetadataPolicy` from what I can see. I'll add a mention of how validation is carried out in the metadata doc.

laneb: Done.

src/main/scala/org/tresamigos/smv/dqm/SmvDQM.scala
laneb: Done.

src/test/python/testSmvFramework.py
cloudysail: Right, just for consistency since we had the requirement to catch `SmvDqmValidationError`. We can do that later as needed.

src/main/scala/org/tresamigos/smv/SmvMetaData.scala
cloudysail: There are `validator_cls.puml` and `validator_seq.puml` mentioned this class. Not sure if we usually do synchronize between puml files and code changes

laneb: Not consistently, but it's a good idea. I've removed those references.

PR #1255, opened 2018-05-29
Enable creating empty projects with 'smv-init'

AntonioRecaldeRusso:
Fixes: #1250 

### Summary
- Allows user to create an empty project using `smv-init` script
- Interface to create projects has not been changed. Creating a simple, test, enterprise projects using `smv-init` remains the same
- New options args added to `smv-init` shell script:
  - `-b | -blank`: creates an empty/blank project
  - `-n | -name`: sets `smv.appName` in `smv-app-conf.props` with the value of the following arg (empty projects only)
  - `-dd | -dataDir`: sets `smv.dataDir` in `smv-user-conf.props` (optional, empty projects only)

E.g.:
- creating a `simple` project (with some sample modules)
```bash
$ ./tools/smv-init /project/path
```
- creating an `empty/blank` project
```bash
$ ./tools/smv-init -b -n <ProjectName> -dd </project/data/dir> </project/path>
```

init

PR #1262, opened 2018-06-01
i1216 fix integration test

### LINKED ISSUE #1216: datasetsForStage sometimes fails due to incorrect result from walk_packages
Need to debug this on the cluster where the problem occurs - it seems that `pkg_util.walk_packages` in some circumstances only returns packages at the shallowest depth. It only seems to occur in cluster mode, which suggests it could be tied to a particular version/installation of Python. The issue seems to be resolved by replacing the following line from `_iter_submodules_in_stage`
```
stagemod = __import__(stage)
```
with
```
stagemod = __import__(stage)
for name in stage.split('.')[1:]:
    stagemod = getattr(stagemod, name)
```

laneb:
Fixes #1216.

### DIFF
--- a/tools/smv-init
+++ b/tools/smv-init
@@ -1,7 +1,6 @@
 #!/bin/bash
 
-TEMPLATE_DIR="$(cd "`dirname "$0"`"; pwd)/templates"
-TEMPLATE_DATA_DIR="$TEMPLATE_DIR/data"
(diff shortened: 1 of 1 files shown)


laneb:
Actually fixed #1255, not #1216 🤦‍♂️

PR #1263, opened 2018-06-01
Use libs in hash (no app props)

### LINKED ISSUE #1247: Incorporate user libraries into SmvDataSet hash

kmannislands:
This is the second PR that could fix #1247 

This approach is different because it removes the need to enumerate libraries in the app props _at all_.

Instead, in `DataSetRepo`, we inspect the modules that we are about to pop from `sys.modules` (which we do in order to force python to load new code for the transaction), check their members for `SmvDataSet`(s) and then load those and call their `requiresLib()` method to assemble the list of libraries we also need to pop from `sys.modules`

### DIFF
--- a/docs/user/app_config.md
+++ b/docs/user/app_config.md
@@ -116,13 +116,6 @@ Note that for sequence/list type parameters (e.g. smv.stages), a "," or ":" can 
 </tr>
 
 <tr>
-<td>smv.user_libraries</td>
-<td>empty</td>
-<td>Optional</td>
-<td>A list of external (not SMV DataSet subclass instances) modules, functions and/or objects that should be reloaded with the project</td>
-</tr>
-
-<tr>
 <th colspan="4">Data Directories Parameters</th>
 </tr>
 
--- a/docs/user/app_config.md
+++ b/docs/user/app_config.md
@@ -116,6 +116,13 @@ Note that for sequence/list type parameters (e.g. smv.stages), a "," or ":" can 
 </tr>
 
 <tr>
+<td>smv.user_libraries</td>
+<td>empty</td>
+<td>Optional</td>
+<td>A list of external (not SMV DataSet subclass instances) modules, functions and/or objects that should be reloaded with the project</td>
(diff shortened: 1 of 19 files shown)


kmannislands:
Would take tips to squeeze additional perf out of this code. Using the `_is_smv_dataset` predicate over `inspect.isclass` helped a lot.

I did a quick informal test from the `smv-pyshell`
```py
import timeit
timeit.timeit('ls()', setup='from smv.smvshell import ls', number=3)
```

Which took ~41s for the props-enumerated approach and ~65s for the inspect-based approach (thats for 3 runs of ls).

One easy optimization would be adding a boolean prop to apps whether or not to do this check, that way legacy apps (or new apps that didn't care about UDL's) could get the old performance? Just a thought.

laneb:
I see a few issues here other than the performance cost:

- This will make it harder to reason about what should/will be reloaded, because the libraries will be declared all across the user's project, and because the reload behavior depends on datasets that are _also_ being reloaded.
- This complicates adding library reloading to an existing project since (at a minimum) the modules which require those libraries will have to be identified
- This makes `_clear_sys_modules` much more complex - there's a lot more surface area for potential bugs, and it will be harder to debug if/when they occur. Since this is already one of the more bug prone areas of the code, I'd be hesitant to complicate it.

Overall, the removal of an item from the conf doesn't seem worth the trade-off in complexity.

If it's important to remove the user_libs from the conf, there's an approach which I think will be faster and simpler: remove every module discovered in `src/main/python` and `library`.

AliTajeldin:
Let's have a whiteboard discussion on this before we start any implementation.

kmannislands:
After discussion with @AliTajeldin and @laneb today, decided on the path-based approach @laneb mentions above. Will close this and complete the work under a new ticket.

PR #1270, opened 2018-06-05
#1269 fix default timestamp format

### LINKED ISSUE #1269: Hour value will be changed to 0 when reading in a timestamp column with hour value is 12
Attached test csv file and schema: [time data.zip](https://github.com/TresAmigosSD/SMV/files/2068072/time.data.zip), smv version 1.6.2 and 2.1.1.3

When column type is `Timestamp` and hour value is `12`, e.g. `2018-02-28 12:31:25.0`, after reading in csv file using `openCsv`, hour value will be changed to 0, e.g. `2018-02-28 00:31:25.0`

See below 3rd and 4th columns:
screen shot 2018-06-04 at 7 55 15 pm

cloudysail:
Fixes #1269. 

Use "yyyy-MM-dd HH:mm:ss.S" as default SMV timestamp format instead of `"yyyy-MM-dd hh:mm:ss.S"`

### DIFF
--- a/docs/user/smv_input.md
+++ b/docs/user/smv_input.md
@@ -212,8 +212,10 @@ other: Decimal[10];
 ### Timestamp type
 The `Timestamp` type can be used to hold a date/timestamp field value.
 An optional format string can be used when defining a field of type `timestamp`.
-The field format is the standard java `java.sql.Timestamp` format string.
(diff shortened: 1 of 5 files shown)


cloudysail:
@laneb docs updated

laneb:
LGTM

PR #1281, opened 2018-06-08
#1280 fix test_smvTimestampToStr

### LINKED ISSUE #1280: Fix test_smvTimestampToStr
The test looks like this
```
    def test_smvTimestampToStr(self):
        df = self.createDF("ts:Timestamp[yyyyMMdd'T'HHmmssZ];tz:String", "20180428T025800+1000,+0000;,America/Los_Angeles;20180428T025800+1000,Australia/Sydney")
        r1 = df.select(col("ts").smvTimestampToStr("+10:00","yyyyMMdd:HHmmssz").alias("localDT"))
        r2 = df.select(col("ts").smvTimestampToStr(col("tz"),"yyyy-MM-dd HH:mm:ssz").alias("localDT2"))

        e1 = self.createDF("localDT: String", "20180428:025800+10:00;;20180428:025800+10:00")
        e2 = self.createDF("localDT2: String", "2018-04-27 16:58:00+00:00;;2018-04-28 02:58:00+10:00")

        self.should_be_same(e1, r1)
        self.should_be_same(e2, r2)
```
We should use `Z`(RFC 822 time zone) rather than `z`(General time zone) according to the [SimpleDateFormat](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html)

cloudysail:
Fixes #1280

### DIFF
--- a/src/test/python/testColumnHelper.py
+++ b/src/test/python/testColumnHelper.py
@@ -187,6 +187,9 @@ class ColumnHelperTest(SmvBaseTest):
 
     def test_smvTimestampToStr(self):
         df = self.createDF("ts:Timestamp[yyyyMMdd'T'HHmmssZ];tz:String", "20180428T025800+1000,+0000;,America/Los_Angeles;20180428T025800+1000,Australia/Sydney")
+        # Use `Z`(RFC 822 time zone) in the SimpleDateFormat because it has only a single valid way to represent a given offset.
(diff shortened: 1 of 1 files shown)


laneb:
Please make a note of why we specifically test the RFC-822 format.

cloudysail:
@laneb updated

PR #1282, opened 2018-06-08
simplify smv-init implementation (#1277)

### LINKED ISSUE #1277: Simplify template copying step smv-init
At its core, `smv-init` should be equivalent to recursively copying a template project and doing a find/replace to update template strings. The step of copying the directory is way too complicated for this goal. Today I tried to add a new directory to the integration test project and ran into problems at multiple steps in `smv-init`.

laneb:
Resolves #1277.


tools/smv-init
```diff
@@ -131,94 +127,27 @@ function create_proj_dir()
...
-
-        cp -R "$data_dir" "$PROJ_DIR"
-    fi
-}
-
-function copy_src_dir()
-{
-    subdir=$1
-    proj_cp=$2
-    DST_DIR="${PROJ_DIR_FULL_PATH}/src/${subdir}/${proj_cp}"
```
cloudysail: In `blank` mode, we need to create a stage folder named as STAGE_NAME under `src/main/python` which is needed in atlas.

tools/smv-init
```diff
@@ -131,94 +127,27 @@ function create_proj_dir()
...
-      done
-    fi
+    local files="$(cd $TEMPLATE_DIR; find . -type f)"
+    local ESCAPED_FULL_PATH=$(echo ${PROJ_DIR_FULL_PATH} | sed -e 's/[\/&]/\\&/g')
+    local ESCAPED_DATA_DIR=$(echo ${PROJ_DATA_DIR} | sed -e 's/[\/&]/\\&/g')
+
+    for f in $files; do
+        sed  -e "s/_GROUP_ID_/$PROJ_GROUP_ID/" \
+             -e "s/_ARTIFACT_ID_/$PROJ_ARTIFACT_ID/" \
+             -e "s/_PROJ_CLASS_/$PROJ_CLASS/g" \
```
cloudysail: remove _PROJ_CLASS_?

laneb: It's still used by the integration test application.

cloudysail: `PROJ_CLASS ` is used in extract_group_artifact_ids() for `test`, while I don't see anywhere we need to do the sed replacement of  `_PROJ_CLASS_` with `$PROJ_CLASS`. Did I miss something?

tools/smv-init
```diff
@@ -245,17 +174,25 @@ function create_git_structure()
     fi
 }
 
+function copy_data_files() {
+    # integration test project has its own data already in the template
+    # other projects share the same data, so it is stored outside of the template
+    if [ "$PROJ_TYPE" != "test" ]; then
```
AntonioRecaldeRusso: some of these checks are inside functions, and some are wrapping the function calls (in the bottom of the file). It should be consisent

tools/smv-init
```diff
@@ -245,17 +174,25 @@ function create_git_structure()
...
+        echo "-- copying template data"
+        cp -r "$TEMPLATE_DATA_DIR" "$PROJ_DIR_FULL_PATH/data"
+    fi
+}
+
 # --- MAIN ---
 
 #  only test proj has PROJ_CLASS
 if [ "$PROJ_TYPE" = "test" ] ; then
```
AntonioRecaldeRusso: I think the if statement should go inside the function

tools/smv-init
```diff
@@ -245,17 +174,25 @@ function create_git_structure()
...
-create_proj_dir
-copy_conf_files
-copy_data_files "$TEMPLATE_DATA_DIR"
-copy_src_files
+copy_proj_dir
+substitute_tokens
+copy_data_files
 create_python_packages
 
 if [ "$PROJ_TYPE" = "big" ] ; then
```
AntonioRecaldeRusso: same

tools/smv-init
laneb: `PROJ_CLASS` determines some other values like the `PROJ_ARTIFACT_ID`, which still exists in the integration project.

tools/smv-init
cloudysail: Still not fixed?

tools/smv-init
```diff
@@ -131,94 +127,27 @@ function create_proj_dir()
...
-      for f in $ENTERPRISE_FILES; do
-          copy_with_inject "$TEMPLATE_DIR/$f" "$PROJ_DIR/$f"
-      done
-    else
-      for f in $PYTHON_FILES; do
-          copy_with_inject "$TEMPLATE_DIR/$f" "$PROJ_DIR/$f"
-      done
-    fi
-
-    # .gitignore was postfixed with .template so it does not affect SMV's git repo
```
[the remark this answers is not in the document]
cloudysail: Still seems an issue?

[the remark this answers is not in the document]
laneb: Done.

tools/smv-init
laneb: Done.

cloudysail:
There are still some issues if you compare the result before and after code changes:
- -b: blank/empty project
`./tools/smv-init -b -n B2 ../new/b1`
There is `data` folder generated. But for `blank project` we should not have the data folder

- -s: simple (default)
`./tools/smv-init ../new/s1`
```
-- creating project directory
-- substituting tokens with project info
sed: 1: "s/_PROJ_NAME_/../new/s1/g
": bad flag in substitute command: 'n'
```

Still need to go through all other project types to see if anytihng goes wrong.

laneb:
@cloudysail I cannot replicate the first one, so I guess it's already fixed. The second one is fixed.

laneb:
Let's attack these in a future ticket. They are outside of the scope of the
original issue and blocking others.

cloudysail:
sounds good.

PR #1287, opened 2018-06-09
 #1045 removed DedupWithOrderGDO

### LINKED ISSUE #1045: smvDedupByKeyWithOrder performance is lower than using window function
Current smvDedupByKeyWithOrder using rdd level operations. On a client dataset, using window function is much faster than our implementation. Should at least make the default behavior of smvDedupByKeyWithOrder using the window function as in 
https://stackoverflow.com/questions/33878370/how-to-select-the-first-row-of-each-group

The order structure approach is also interesting, although it only work for selected columns.

ninjapapa:
Fixed #1045

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SmvDFHelper.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvDFHelper.scala
@@ -505,10 +505,11 @@ class SmvDFHelper(df: DataFrame) {
 
   /** Same as `dedupByKeyWithOrder(Column*)(Column*)` but use `String` as key **/
   def dedupByKeyWithOrder(k1: String, krest: String*)(orderCol: Column*): DataFrame = {
-    val gdo = new cds.DedupWithOrderGDO(orderCol.map { o =>
-      o.toExpr
-    }.toList)
-    df.smvGroupBy(k1, krest: _*).smvMapGroup(gdo).toDF
+    val w = Window.partitionBy(k1, krest: _*).orderBy(orderCol: _*)
+    val tmpCol = mkUniq(df.columns, "_dedupByKeyWithOrder_rank", ignoreCase = true, "_")
+    df.withColumn(tmpCol, row_number().over(w))
+      .where(col(tmpCol) === 1)
+      .drop(tmpCol)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/main/scala/org/tresamigos/smv/SmvDFHelper.scala
```diff
@@ -505,10 +505,11 @@ class SmvDFHelper(df: DataFrame) {
 
   /** Same as `dedupByKeyWithOrder(Column*)(Column*)` but use `String` as key **/
   def dedupByKeyWithOrder(k1: String, krest: String*)(orderCol: Column*): DataFrame = {
-    val gdo = new cds.DedupWithOrderGDO(orderCol.map { o =>
-      o.toExpr
-    }.toList)
-    df.smvGroupBy(k1, krest: _*).smvMapGroup(gdo).toDF
+    val w = Window.partitionBy(k1, krest: _*).orderBy(orderCol: _*)
+    val tmpCol = mkUniq(df.columns, "_dedupByKeyWithOrder_rank", ignoreCase = true, "_")
```
laneb: No need for `tmpCol` to begin with:
```
    val w = Window.partitionBy(k1, krest: _*).orderBy(orderCol: _*)
    df.where(row_number().over(w) === 1)
```

tools/templates/test/build.sbt
```diff
@@ -13,7 +13,7 @@ val sparkVersion = "2.3.0"
 libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-sql"  % sparkVersion % "provided",
   "org.apache.spark" %% "spark-hive" % sparkVersion % "provided",
-  "org.tresamigos"   %% "smv"        % "2.2-SNAPSHOT",
+  "org.tresamigos"   %% "smv"        % "2.3-SNAPSHOT",
```
laneb: I need to make a list of all of the places I need to apply the update...

src/main/scala/org/tresamigos/smv/SmvDFHelper.scala
laneb: Actually, this compiles but fails with a strange errors. Seems to be this bug: https://issues.apache.org/jira/browse/SPARK-16418

PR #1288, opened 2018-06-11
I1267 smv driver interface

### LINKED ISSUE #1267: SMV should distinguish between driver script args and SMV args
All args following `--script <script name>` and preceding any `--` should be treated as arguments to the driver script. Arguments preceding `--script` should be treated as SMV args. It is prohibitive to expect the user to distinguish between the two at runtime.

laneb:
Resolves #1267.

@jacobdr fyi


src/main/python/scripts/runapp.py
```diff
@@ -11,9 +11,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from smv import SmvApp
 import sys
 
-app = SmvApp.createInstance(sys.argv[1:])
-# skip the first argument, which is this program
-app.run()
+from smv.smvdriver import SmvDriver
```
AliTajeldin: I don't think the `import sys` is needed any more.

src/main/python/smv/smvdriver.py
```diff
@@ -0,0 +1,33 @@
...
+from smv import SmvApp
+
+class SmvDriver(object):
+    """Driver for an SMV application
+
+        SmvDriver handles the boiler plate around parsing driver args, constructing an SmvApp, and running an application.
+        To use SmvDriver, override `main` and in the main block of your driver script call construct your driver and
+        call `run`.
+    """
+    def create_smv_app(self, smv_args, driver_args):
```
AliTajeldin: Should document this function as it is for the user to override if they want to create a custom smv app that is dependent on the driver args.

src/main/python/smv/smvdriver.py
```diff
@@ -0,0 +1,33 @@
...
+        return SmvApp.createInstance(smv_args)
+
+    def main(self, app, driver_args):
```
AliTajeldin: Should document this as this is the function that the user would normally override.

laneb:
Added all requested documentation.

PR #1290, opened 2018-06-11
I1271 from 2.3 rm gdo

### LINKED ISSUE #1271: Remove SmvGDO totally
As part of Smv history, SmvGDO was designed to iterate through all the records in a group by some keys. This approach was introduced when there are no easy native Spark function to address the need.  `SmvGDO` used some Spark internal implementation details which were not supported anymore. 

We need to finally remove `SmvGDO` totally and use native Spark functions to support the used cases.

Currently, the following GDOs are in the code:
* `SmvQuantile` - dead code, can be removed already (since we used window function now)
* `SmvChunkUDF` and `SmvChunkUDFGDO` - never ported `chunkBy` method to Python, so can drop support, so can remove also
* `DedupWithOrderGDO` - issue #1045 will remove the need for this GDO
* `FillNullWithPrev` - is in use and has Python side api dependency 

For the `FillNullWithPrev`, we should figure out a Spark native approach to implement.

ninjapapa:
fixed #1271 

SmvGDO is gone!


laneb:
lgtm

PR #1298, opened 2018-06-13
i1186 runconfig less aggressive influence on dataset hash

### LINKED ISSUE #1186: SmvRunConfig is too aggressive about changing the dataset hash
Expected behavior:  hash changes when a key in use by the dataset changes

Observed behavior: hash changes when _any_ key changes

laneb:
Resolves #1186.

@jacobdr @chengfei0913 fyi


docs/user/run_config.md
```diff
@@ -13,17 +13,18 @@ smv.config.filetype=csv
...
-    return df
+  def run(self, i):
+    ...
+    if self.smvGetRunConfig("sample"):
+      res = res.sample(0.01)
+    return res
 ```
 
-The module hash value of modules that inherit from `SmvRunConfig` will be affected by the current config values.  The hash will utilize **ALL** the config values even ones that are not used by the current module.
+The module hash value of modules that inherit from `SmvRunConfig` will be affected by **ALL** config values - even ones that are not used by the current module.
```
ninjapapa: Is this part of the doc accurate?

src/main/python/smv/smvdataset.py
```diff
@@ -98,8 +128,15 @@ def requiresDS(self):
             Returns:
                 (list(SmvDataSet)): a list of dependencies
         """
+        pass
+
+    def requiresConfig(self):
```
ninjapapa: Should we have this interface documented in the user doc?

docs/user/run_config.md
```diff
@@ -13,17 +13,18 @@ smv.config.filetype=csv
...
 ```python
 class MyModule(SmvModule, SmvRunConfig):
-  def run():
-    if self.smvGetRunConfig("sample") == "1pct":
-      df = df.sample(0.01)
-    return df
+  def run(self, i):
+    ...
+    if self.smvGetRunConfig("sample"):
+      res = res.sample(0.01)
```
cloudysail: Is `res` the current df? May need to add a comment to avoid confusion

docs/user/run_config.md
laneb: 85% true. Actually, it will be affected by all of keys declared in `smv.config.keys` and their values. So if a config value affects one module's hash, it affects every module's hash. Will update the doc to be more specific.

src/main/python/smv/smvdataset.py
```diff
@@ -164,19 +201,27 @@ def sourceCodeHash(self):
...
                     res += m(self.smvApp).sourceCodeHash()
-            except: pass  # noqa: E701
+            except: 
+                pass
+
 
         # if module inherits from SmvRunConfig, then add hash of all config values to module hash
-        if hasattr(self, "_smvGetRunConfigHash"):
+        try:
             res += self._smvGetRunConfigHash()
```
cloudysail: Since we already have `self.config_hash()`, shall we remove these?

src/main/python/smv/smvdataset.py
laneb: Yes, I've added a note about it.

src/main/python/smv/smvdataset.py
laneb: That would break existing projects in the sense that the config would have no influence on existing modules that mixin `SmvRunConfig`. Better to leave that behavior as is for now. So the new recommended pattern is to use `requiresConfig`, but `SmvRunConfig` will still influence the hash as before. Really, that's the only thing it'll do. I'll add a couple of notes to explain this.

ninjapapa:
Nice!

cloudysail:
@laneb `test_run_module_with_dynamic_run_config` is failing in travis. Something needs to fix?

laneb:
Thought I'd already fixed that one. One of the examples just needs to be updated to use `requiresConfig`.

PR #1306, opened 2018-07-06
I1305: add method to get schema from CSV file path

### LINKED ISSUE #1305: Provide method to discover schema of CSV files
Task is to create a method on the `smvApp` to get the schema of a CSV file from its path

chengfei0913:
Fixes #1305

### DIFF
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -225,6 +225,7 @@ class SmvApp(object):
 
     def getSchemaByDataFileAsSmvSchema(self, data_file_name):
         """Get the schema of a data file from its path and returns a Scala SmvSchema instance.
+           The result will be None if the corresponding schema file does not exist or is invalid.
         """
         data_file_path = os.path.join(self.inputDir(), data_file_name)
         return self.j_smvPyClient.readSchemaFromDataPathAsSmvSchema(data_file_path)
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -226,12 +226,8 @@ class SmvApp(object):
     def getSchemaByDataFileAsSmvSchema(self, data_file_name):
         """Get the schema of a data file from its path and returns a Scala SmvSchema instance.
         """
-        smv_schema_obj = self.j_smvPyClient.getSmvSchema()
         data_file_path = os.path.join(self.inputDir(), data_file_name)
-        schema_file_path = smv_schema_obj.dataPathToSchemaPath(data_file_path)
-        smv_schema_instance = smv_schema_obj.fromFile(self.j_smvApp.sc(), schema_file_path)
-
-        return smv_schema_instance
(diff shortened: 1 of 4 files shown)


src/main/python/smv/smvapp.py
```diff
@@ -223,6 +223,18 @@ def discoverSchemaAsSmvSchema(self, path, csvAttributes, n=100000):
...
+    def getSchemaByDataFileAsSmvSchema(self, data_file_name):
+        """Get the schema of a data file from its path and returns a Scala SmvSchema instance.
+        If no corresponding schema file, then returns None.
+        """
+        smv_schema_obj = self.j_smvPyClient.getSmvSchema()
+        data_file_path = os.path.join(self.inputDir(), data_file_name)
+        schema_file_path = smv_schema_obj.dataPathToSchemaPath(data_file_path)
+        try:
+            return smv_schema_obj.fromFile(self.j_smvApp.sc(), schema_file_path)
+        except:
```
laneb: catch-all except is sketchy. Let's push this logic into the pyclient and handle the exact error that we expect there. If you return `null` from scala, py4j will translate it to `None`.

chengfei0913: File does not exist will cause an `InvalidInputException`.

@lane, Do you mean that we should catch `InvalidInputException` in `fromFile` at scala side?
If yes, then we will return `null` for `InvalidInputException`, but throw all other exceptions/error.
Does this make sense?

And we need to change the logic of `isPersisted` in `SmvDataSet.scala`
```scala
  /** Has the result of this data set been persisted? */
  private[smv] def isPersisted: Boolean =
    Try(new FileIOHandler(app.sqlContext, moduleCsvPath()).readSchema()).isSuccess
```
Otherwise, it will return `true` even schema file does not exist.

laneb: We don't need to push it all the way down into `fromFile`. Let's not change `SmvSchema` at all. If it was feasible to catch the exception in the Python `getSchemaByDataFileAsSmvSchema` we would do it there, but the error will be a generic `py4j` error. So we just need to create an adapter in Scala to handle the error before returning the result. Create an `SmvPyClient` method for getting the schema. You can catch the exception there.

chengfei0913:
@laneb Please re-review

src/main/python/smv/smvapp.py
```diff
@@ -223,6 +223,12 @@ def discoverSchemaAsSmvSchema(self, path, csvAttributes, n=100000):
         """
         return self._jvm.SmvPythonHelper.discoverSchemaAsSmvSchema(path, n, csvAttributes)
 
+    def getSchemaByDataFileAsSmvSchema(self, data_file_name):
+        """Get the schema of a data file from its path and returns a Scala SmvSchema instance.
```
laneb: Add a note that the result will be None if the schema doesn't exist or is invalid.

chengfei0913: Added

chengfei0913:
@laneb Comment and tests are added

laneb:
Looks good!

Issue #988, opened 2017-09-27
Improvement on smvHist function


yw-yang:
When input column is a numerical type, smvHist prints histogram with defaulted bin size 100.

`train_new.smvHist("binary_target")`

> Histogram of binary_target: with BIN size 100.0
> key                      count      Pct    cumCount   cumPct
> 0.0                      32561  100.00%       32561  100.00%

This behavior is not quite user friendly. If users want to do a distribution check on a column with limited numbers, they have to pick up an appropriate bin size and use smvBinHist like below, which is a little complex.

`train_new.smvBinHist(("binary_target",1))`

> Histogram of binary_target: with BIN size 1.0
> key                      count      Pct    cumCount   cumPct
> 0.0                      24720   75.92%       24720   75.92%
> 1.0                       7841   24.08%       32561  100.00%

Could we just improve the function and print all values in this column? I understand # of vales will be pretty large in some cases but users can always choose smvBinHist to limit the output.

ninjapapa:
Need a little design here.

We can not make the histogram on a numerical field without binning too easy to trigger, especially when user does not even know what the distribution may look like. It will easily hang the system and run out of memory. 

However we could introduce a method which named clearly so that no one will use it unintentionally. `smvNumValueHist` sounds too long and still unclear. @yw-yang any suggestions?

yw-yang:
My opinion:
1. The defaulted behavior for numerical column in `smvHist` function is confusing. User may ignore the bin size 100 unintentionally and take the wrong result. I suggest to change the behavior.
2. Introduce another method `smvNumValueHist` seems a little duplicated with `smvHist` and also a burden for user to remember another hist function.
3. The concern that too many values may hang the system will also appear in string field(e.g. do histogram on PNR or name). So is it possible we set a threshold internally, e.g. if # values is greater than 1000, method stops calculating and shows a warning instead.

ninjapapa:
Good idea. Let's take approach 3 as a brief design and refine smvHist.

yw-yang:
This ticket is also related to the visualization improvement in dsalib. Please refer to design below:

**Design:**
1. Add another two EDD task types: `NumericByKeyHistogram` and `NumericByFreqHistogram`
2. Change class `Hist` default bin size from 100 to 0
3. Change Edd logic for numeric types as below
```
      case _: NumericType => 
        if (histCol.binSize > 0.0) edd.BinNumericHistogram(s, histCol.binSize)
[...]

PR #1313, opened 2018-07-19
i1312: Support date and timestamp types in smv histogram helper functions

### LINKED ISSUE #1312: Allow smv histogram functions to support date and timestamp types
Currently smv histogram functions only allow string, numeric and boolean types. We should also support date and timestamp types.

```
  private def createHist(histCol: edd.Hist) = {
    val s = df(histCol.colName)
    df.schema(histCol.colName).dataType match {
      case StringType =>
        if (histCol.sortByFreq) edd.StringByFreqHistogram(s)
        else edd.StringByKeyHistogram(s)
      case _: NumericType => edd.BinNumericHistogram(s, histCol.binSize)
      case BooleanType    => edd.BooleanHistogram(s)
      case t              => throw new SmvUnsupportedType(s"data type: ${t} is not supported")
    }
  }
```

yw-yang:
Fix #1312 

screen shot 2018-07-19 at 1 50 14 pm

### DIFF
--- a/src/main/scala/org/tresamigos/smv/edd/EddTask.scala
+++ b/src/main/scala/org/tresamigos/smv/edd/EddTask.scala
@@ -203,27 +203,15 @@ private[smv] case class BooleanHistogram(col: Column) extends EddHistTask {
   override val statOp   = histBoolean(col)
 }
 
-private[smv] case class DateHistogram(col: Column) extends EddHistTask {
-  override val taskName = "dat"
-  override val taskDesc = "Date"
-  override val statOp   = histStr(col.cast("string"))
-}
-
-private[smv] case class TimeHistogram(col: Column) extends EddHistTask {
-  override val taskName = "tim"
-  override val taskDesc = "Timestamp"
-  override val statOp   = histStr(col.cast("string"))
-}
-
 private[smv] case class StringByKeyHistogram(col: Column) extends EddHistTask {
   override val taskName = "key"
   override val taskDesc = "String sort by Key"
-  override val statOp   = histStr(col)
+  override val statOp   = histStr(col.cast("string"))
 }
 
 private[smv] case class StringByFreqHistogram(col: Column) extends EddHistTask {
   override val taskName   = "frq"
   override val taskDesc   = "String sorted by Frequency"
   override def sortByFreq = true
-  override val statOp     = histStr(col)
+  override val statOp     = histStr(col.cast("string"))
 }
--- a/src/main/scala/org/tresamigos/smv/edd/EddTask.scala
+++ b/src/main/scala/org/tresamigos/smv/edd/EddTask.scala
@@ -206,6 +206,8 @@ private[smv] case class BooleanHistogram(col: Column) extends EddHistTask {
 private[smv] case class StringByKeyHistogram(col: Column) extends EddHistTask {
   override val taskName = "key"
   override val taskDesc = "String sort by Key"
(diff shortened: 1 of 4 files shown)


laneb:
@yw-yang Is your git configured with your github email address? Your commits are stamped with your name but no email or avatar.

yw-yang:
@laneb I just made some changes, are you able to see email or avatar now?

src/main/scala/org/tresamigos/smv/edd/EddTask.scala
```diff
@@ -206,12 +206,12 @@ private[smv] case class BooleanHistogram(col: Column) extends EddHistTask {
 private[smv] case class StringByKeyHistogram(col: Column) extends EddHistTask {
   override val taskName = "key"
   override val taskDesc = "String sort by Key"
-  override val statOp   = histStr(col)
+  override val statOp   = histStr(col.cast("string"))
```
laneb: Needs documentation.

laneb:
@yw-yang No, but if you changed your configuration we should see the change when you push your next commit.

src/main/scala/org/tresamigos/smv/edd/EddTask.scala
yw-yang: Comments have been added.

yw-yang: I agree doing a histogram of timestamp is rare and not very useful but still think smv should support it for the rare case and let user decide whether to use it or now.

yw-yang:
I agree doing a histogram of timestamp is rare and not very useful but still think smv should support it for the rare case and let user decide whether to use it or now.

laneb:
👍 
yuwei commit avatar

laneb:
The question was whether it makes sense to do a histogram on the raw timestamp or whether the timestamps ought to be binned into larger time intervals similar to the `BinNumericHistogram`. I suppose. Does the timestamp hist feature have no use case?

yw-yang:
@laneb I rarely do histogram on timestamp type, the use case I encountered before is to check the `system_load_ts`, which records the timestamp each row is appended to a table. In batch process, for each batch loading, the `system_load_ts` is always the same, so there won't be a large number of different values. Another case is in some ETL systems, although `system_load_ts` is timestamp type, the value just contains date information, e.g. "2018-01-01 00:00:00". Of course, user can always cast timestamp type to date/string type and do `smvHist`, but it is an additional burden and not user-friendly. 

Another concern is doing histogram on timestamp may hang the system and run out of memory. In ticket #988,  I propose we can set a threshold in `smvHist` internally, e.g. if # values is greater than 1000, method stops calculating and shows a warning instead. What do you think?

laneb:
That makes sense. Maybe we should make the timestamp binning usecase a little more explicit? If we are expecting that timestamps are binned in order for this to work, we should probably make the binning explicit and a bit easier - e.g. enumerate a few levels of granularity that the timestamps can be binned into. Then we can also make the representation in the UI a bit nicer - truncate the timestamp string to the significant digits.

yw-yang:
Do you mean creating another function, e.g. `smvTimeHist(*colwithGranularity)` to deal with timestamp binning like below? If so, I will open another ticket for it.
```
d.smvTimeHist(("ts1", "Year"), ("ts2", "Month"))
```

laneb:
Yeah, it's outside of the scope of this ticket. Just seems like it would be a more useful API if we really want to support timestamp hist.

PR #1320, opened 2018-07-26
I1308 docker base image for release

### LINKED ISSUE #1308: Bring back SMV base Docker image for building release
Currently we have a bit of a convoluted approach to using Docker to standardize the environment for the release build. The short of it is that we build a docker image based on the old release to use when building the new release. This leads to complications. Instead we should create a base image with all of SMV's dependencies to use when building the new release artifact. The SMV image which we publish will be based on this image and will install SMV by unpacking the release artifact which we've just published.

laneb:
Resolves #1308.

@jacobdr Please have a look tomorrow if you have time.

The goal of this PR is to reduce the complexity of `smv-release`.
1. Created a multi-stage docker build. First stage installs dev dependencies and generates the release artifacts. Second stage install prod dependencies and unpacks the release artifact.
2. Created make targets to create both docker images, extract the release artifact from the first, and build docs
3. Updated smv-release to leverage these make targets


docker/smv/Dockerfile
```diff
@@ -1,63 +1,75 @@
...
-    mkdir ${TEMPLATE_DIR}
+ENV TEMPLATE_DIR=/home/template
 
+# TODO: does the docker image really need git or r-base any more?
 RUN echo "deb http://cran.rstudio.com/bin/linux/debian lenny-cran/" >> /etc/apt/source.list &&\
     apt-key adv --keyserver hkp://pgp.mit.edu --recv-key 381BA480 &&\
     apt-get update &&\
-    apt-get install -y r-base &&\
-    apt-get install -y unzip &&\
+    apt-get install -y r-base &&\
```
AliTajeldin: I thought we are removing this.  Or is that a separate issue?

docker/smv/Dockerfile
```diff
@@ -1,63 +1,75 @@
...
+ENV PATH=${SPARK_HOME}/bin:${SMV_HOME}/tools:${PYENV_ROOT}/bin/:${PATH}
+
+# create the projects directory and create a flag (.docker) to indicate project is inside docker image.
+RUN mkdir /projects &&\
+    touch /projects/.docker &&\
+    mkdir ${TEMPLATE_DIR}
+
+COPY --from=smv-build ${PYENV_ROOT} ${PYENV_ROOT}
+COPY --from=smv-build /usr/lib/SMV/smv_*.tgz .
+RUN tar xzvf smv_*.tgz && rm -rf smv_*.tgz && mv ./SMV ${SMV_HOME} &&\
```
AliTajeldin: This block really needs a comment to explain what is going on and why.

docker/smv/Dockerfile
laneb: There's a separate issue for removing R

docker/smv/Dockerfile
laneb: Added a header comment to describe the multi-stage pattern and some inline comments explaining the copies from first to second stage.

Makefile
```diff
@@ -25,6 +34,68 @@ publish-scala: assemble-fat-jar
...
+
+PYDOC_DEST := $(DOC_DIR)/python
+SMV_PY_DIR := $(SMV_ROOT)/src/main/python
+SPARK_PY_DIR := $(DEFAULT_SPARK_HOME)/python
+PY4J_LIBS = $(shell find $(SPARK_PY_DIR)/lib -maxdepth 1 -mindepth 1 -name 'py4j*-src.zip' -print | tr -d '\r')
+
+py-doc: $(PYDOC_DEST)
+
+$(PYDOC_DEST): install-spark-default
+	env SMV_VERSION=$(SMV_VERSION) \
```
laneb: I doubt using `env` is the best solution to the problem of exporting env variables to sphinx. It was the first I got to work. Need to at least dry out this approach since it is c/p'd a few lines down.

docker/smv/Dockerfile
```diff
@@ -61,7 +76,13 @@ RUN mkdir /projects &&\
     touch /projects/.docker &&\
     mkdir ${TEMPLATE_DIR}
 
+# Copy the pyenv installation from the first stage. This is an optimization so we don't install Python twice.
+# we can get away with this because we don't currently differentiate between dev and prod requirements.
 COPY --from=smv-build ${PYENV_ROOT} ${PYENV_ROOT}
+# Copy the SMV release artifact from the first stage and unpack it. Since the same artifact is also copied
```
AliTajeldin: nice!

Makefile
AliTajeldin: actually `env` is the "unixish" way.  That is the whole purpose of `env` existence 😄

PR #1326, opened 2018-09-11
I1324 spark dir env var

### LINKED ISSUE #1324: smv-install can't find SPARK_DIR
Since `smv-install` call `spark-install` to install spark, itself does not know whether spark get installed, so variable `SPARK_DIR` used in `smv-intall` to create `smv-profile.sh` is aways empty. 

As a result, when run smv, it can't find spark-submit.

ninjapapa:
Fixed #1324 

Instead of let `spark-install` to specify the `SPARK_DIR`, let `smv-install` to specify it as `Spark`. 

Tested by:
```
curl https://raw.githubusercontent.com/TresAmigosSD/SMV/i1324_spark_dir_env_var/tools/smv-install | bash -s -- -spark
```

### DIFF
--- a/tools/smv-install
+++ b/tools/smv-install
@@ -38,6 +38,8 @@ function parse_args()
   SMV_PROFILE_FILENAME="smv-profile.sh"
   SMV_PROFILE="${HOME}/.smv/${SMV_PROFILE_FILENAME}"
 
+  SPARK_DIR="Spark"
+
   validate_version
 }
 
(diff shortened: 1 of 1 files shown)


jacobdr:
I know this got merged already, but do we need to worry about the case sensitivity at all across systems? @ninjapapa

ninjapapa:
@jacobdr Could you explain? I'm not so sure I understand the use case.

PR #1334, opened 2018-09-18
I1284 remove SmvGit

### LINKED ISSUE #1284: Remove SmvGit

tianruiwang:
Fixes #1284 
Removed class `SmvGit` and its two interfaces from SMV.


laneb:
Should also remove the dependency on jgit from build.sbt.

tianruiwang:
Good catch! Dependency removed.

AliTajeldin:
Will wait for official 👍 from @laneb before merge.

PR #1337, opened 2018-09-18
Add 1st column value as column description to schemaDiscovery

### LINKED ISSUE #74: Add one sample data in schema for schemaDiscovery
Instead of generate schema file as

```
laim_gid: Long
srvc_line_seq: Integer
admin: Integer
PATIENT_GID: Integer
```

Add a single data example as comment for each field as

```
laim_gid: Long          #00018231
srvc_line_seq: Integer      #1267
admin: Integer                  #1
PATIENT_GID: Integer    #00991123
```

ninjapapa:
Fixed #74 

* Split original `SmvKeys.scala` file to `SmvKeys.scala` and `SchemaMetaOps.scala` files
* Made `SmvKeys` an object for all the utility methods used for StructField metadata operations
* Added first recode in SchemaDiscovery process to be the SmvDesc for the discovered Schema file

A result:
```
@delimiter = ,
@has-header = true
@quote-char = "
id: Integer @metadata={"smvDesc":"1"}
name: String @metadata={"smvDesc":"Scott"}
age: Integer @metadata={"smvDesc":"19"}
weight: Float @metadata={"smvDesc":"160.98"}
active: Boolean @metadata={"smvDesc":"true"}
address: String @metadata={"smvDesc":""}
registration_date: Date[yyyy-MM-dd] @metadata={"smvDesc":"2012-08-03"}
last_active_date: Date[yyyyMMdd] @metadata={"smvDesc":"20140930"}
last_active_time: Timestamp[yyyy-MM-dd HH:mm:ss] @metadata={"smvDesc":"2014-09-03 14:36:15"}
```

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SchemaDiscoveryHelper.scala
+++ b/src/main/scala/org/tresamigos/smv/SchemaDiscoveryHelper.scala
@@ -152,15 +152,16 @@ class SchemaDiscoveryHelper(sqlContext: SQLContext) {
     // An alternative is to use the mapPartitionWithIndex
     val rowsToParse = noHeadRDD.take(numLines)
 
-    val columnsWithIndex = columns.zipWithIndex
-
     var validCount = 0
+    //abstract first valid record to be used in the discovered schema as the columne description
+    var firstValidRec = new Array[String](columns.length)
     for (rowStr <- rowsToParse) {
       val rowValues = Try { parser.parseLine(rowStr) }.getOrElse(Array[String]())
-      if (rowValues.length == columnsWithIndex.length) {
+      if (rowValues.length == columns.length) {
         validCount += 1
         for (index <- 0 until columns.length) {
           val colVal = rowValues(index)
+          if (validCount == 1) firstValidRec(index) = colVal
           if (colVal.nonEmpty) {
             typeFmts(index) = getTypeFormat(typeFmts(index), colVal)
           }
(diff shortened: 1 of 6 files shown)


src/main/scala/org/tresamigos/smv/SchemaDiscoveryHelper.scala
```diff
@@ -152,15 +152,16 @@ class SchemaDiscoveryHelper(sqlContext: SQLContext) {
...
+    //extract first valid record to be used in the discovered schema as the columne description
+    var firstValidRec = new Array[String](columns.length)
     for (rowStr <- rowsToParse) {
       val rowValues = Try { parser.parseLine(rowStr) }.getOrElse(Array[String]())
-      if (rowValues.length == columnsWithIndex.length) {
+      if (rowValues.length == columns.length) {
         validCount += 1
         for (index <- 0 until columns.length) {
           val colVal = rowValues(index)
+          if (validCount == 1) firstValidRec(index) = colVal
```
AliTajeldin: We should wait to add `colVal` as desc only if `colVal.nonEmpty`.  That is, move this line inside the conditional.  That way, if the first line had nulls, we would avoid "nulls" in the description and use the first non-null value.
May need to rename `firstValidRec` in that case.

ninjapapa: Make sense. Will do.

Issue #1339, opened 2018-09-18
Remove Scala SmvDataSet interface classes


ninjapapa:
Keep the `SmvDataSet` base class and all necessary classes for Python side modules continue to work as is. Remove every other classes derived from `SmvDataSet` on Scala side.

ninjapapa:
Will do this in multiple steps: 

1. Remove the standalone modules, such as `SmvHiveTable`, `SmvMultiCsvFiles` etc.
2. Remove `SmvCsvFile`
3. Clean up

PR #1344, opened 2018-09-19
I1339 remove scala modules

ninjapapa:
Step1 of #1339 

* Removed python shell side dependency to scala modules
* Removed SmvMultiCsvFiles, SmvFrlFile, SmvHiveTable


AliTajeldin:
neato!

PR #1345, opened 2018-09-20
I1339 part2 remove scala smvcsvfile

ninjapapa:
Step2 of #1339 

Removed SmvCsvFile & SmvFile from Scala side.


tools/templates/test/conf/smv-app-conf.props
```diff
@@ -2,16 +2,10 @@
 smv.appName = Example App
 
 # stage definitions
-smv.stages = integration.test.test1, \
-             integration.test.test2, \
```
laneb: 😢

PR #1349, opened 2018-09-20
I1339 step3 cleanup test

### LINKED ISSUE #1339: Remove Scala SmvDataSet interface classes

ninjapapa:
Fixed #1339 

Removed the following from Scala side:
* `Using[SmvRunConfig]`
* `SmvAncillary`/`SmvHierarchy` 
* `SmvJdbcTable`
* scala shell commands
* all scala integration tests


src/main/scala/org/tresamigos/smv/SmvDataSet.scala
```diff
@@ -710,42 +694,6 @@ private[smv] abstract class SmvInputDataSet extends SmvDataSet {
   def run(df: DataFrame) = df
 }
 
-/**
- * Wrapper for a database table accessed via JDBC
- */
-class SmvJdbcTable(override val tableName: String)
```
AliTajeldin: Are we dropping support for jdbc input?  Should we wait to remove this after we have the python equivalent?

laneb: We have a Python implementation based on the DataFrameReader. Should be ready to drop the Scala implementation.

ninjapapa: https://github.com/TresAmigosSD/SMV/blob/i1339_step3_cleanup_test/src/main/python/smv/smvinput.py#L404

AliTajeldin: cool beans.

PR #1350, opened 2018-09-21
I1348 rm dqm udpolicy

### LINKED ISSUE #1348: Remove DQM user defined policy dependency to DF
Current DQM supports user defined policy on Scala side:

```scala
val policy: (DataFrame, DQMState) => Boolean = {(df, state) =>
  val avgPrice = df.agg(avg($"Price")).first.toSeq.head.asInstanceOf[Double]
  avgPrice < 100.0
}
```

This interface allow user to create policy on the entire output DF. 

Since metadata framework in general allows user to operate on the output DF, it is unnecessary to support it from DQM interface anymore.

By dropping this interface, DQM interface will only apply policy on the `DQMState`. It also make it possible to delay DQM policy check to whatever action take place, so we can drop the complicated force action check.

ninjapapa:
Fixed #1348 

Instead of remove the entire user defined policy, simply made it only take DQMState as input. Just remove the df as an input. It solves the problem but keeps the flexibility.

### DIFF
--- a/docs/user/dqm.md
+++ b/docs/user/dqm.md
@@ -224,29 +224,13 @@ override def dqm() = SmvDQM().
     add(DQMRule($"Price" < 1000000.0, "rule1")).
     add(DQMRule($"Price" > 0.0, "rule2")).
     add(DQMFix($"age" > 120, lit(120) as "age", "fix1")).
-    add(policy, "my_udp1")
+    add(DQMPolicy(policy, "my_udp1"))
 ```
 
 **Python**
 ```python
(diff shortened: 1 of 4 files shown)


AliTajeldin:
@ninjapapa is this just the first part?  Is this just removing the DF parameter from policy and not ability to create UD policy?

ninjapapa:
@AliTajeldin I changed my mind :-) 
No need to delete the entire UD policy, removing DF access from UD policy is good enough. I modified the issue itself.

PR #1355, opened 2018-09-21
I1341 move smvDesc and smvLabel to python

### LINKED ISSUE #1341: Move SmvDesc and SmvLabel to Python side
We have `SmvKeys` concept to put/get Smv specific column level info to the `metadata` element of `StructField` of the column. There are 2 places used this concept:
* DF helpers which add/get/delete smvDesc and smvLabel
* Schema discovery add first value to smvDesc

This issue will not move the schema discovery logic to python, so the scope of this issue is to move most of the `SmvKeys` helper, entire `SchemaMetaOps` and all desc and label related DF helpers to python side. We will not support desc and label on Scala DF any more.

However we may need to leave some SmvKeys helpers within Scala to support Schema Discovery for now. Will remove that part when removing Schema Discovery from Scala.

tianruiwang:
Fixes #1341 
Moved all smvDesc, smvLabel related helper functions and tests to python side.

Some comments:
- in order to propagate the metadata of a column in the selecting/renaming/etc. processes, we somehow need to use the DataFrame's "select with metadata" method (e.g. `df.select(col("a").alias("b", meta = {...}))`), which is only supported by pyspark 2.2 or higher versions. As we still need to support pyspark 2.1 for now, this logic is currently implemented on the scala smv side.


src/main/python/smv/schema_meta_ops.py
```diff
@@ -0,0 +1,137 @@
...
+    Provides helper functions for SmvDesc and SmvLabel operations
+"""
+import json
+from pyspark.sql import DataFrame
+from smv.error import SmvRuntimeError
+
+smv_label = "smvLabel"
+smv_desc = "smvDesc"
+
+def getMetaDesc(m):
```
laneb: What's an `m`?

src/main/python/smv/schema_meta_ops.py
```diff
@@ -0,0 +1,137 @@
...
+        self._jPythonHelper = df._sc._jvm.SmvPythonHelper
+
+    def getMetaByName(self, colName):
+        """Returns the metadata of the first column that matches the column name
+
+            Will throw if there's no column that matches the specified name
+        """
+        try:
+            meta = next(col.metadata for col in self.df.schema.fields if col.name == colName)
+        except:
```
laneb: Should only except `StopIteration`

src/main/scala/org/tresamigos/smv/SmvKeys.scala
```diff
@@ -19,41 +19,14 @@ import org.apache.spark.sql.types.{Metadata, MetadataBuilder, StructField}
 /** Utility for using the metadata part of StructField to store Smv keys 
  */
 private[smv] object SmvKeys {
```
ninjapapa: Is this object still needed?

docs/user/smv_desc_label.md
```diff
@@ -90,21 +84,18 @@ When the persisted data read back in, the descriptions and labels are still ther
...
-
+```
+>>> from pyspark.sql.functions import col
+>>> df = app.createDF("a:String", "a")
+>>> df1 = df.smvDesc(("a", "test col a"))
+>>> df1.select(col("a").alias("b")).smvGetDesc()
+[("b", "test col a")]
 ```
 
 As you can see, rename will persist the metadata when you referring the original
```
ninjapapa: No need for the "when ..." part.

src/main/python/smv/schema_meta_ops.py
```diff
@@ -0,0 +1,137 @@
...
+            raise SmvRuntimeError("column name {} not found".format(colName))
+        return meta
+
+    def getDesc(self, colName):
```
laneb: Need docstrings for all of these.

src/main/python/smv/schema_meta_ops.py
ninjapapa: Should we put underscore in front to make those functions "private"?

src/main/scala/org/tresamigos/smv/SmvDFHelper.scala
```diff
@@ -1035,144 +1035,18 @@ class SmvDFHelper(df: DataFrame) {
...
-   * scala> val res = df.smvDesc("a" -> "column a is ....")
-   * scala> res.printDesc
-   * }}}
-   **/
-  def printDesc() = {
-    val discs = df.smvGetDesc().toMap
-    val width = discs.keys.map { _.size }.max
-    discs.foreach { case (n, d) => printf(s"%-${width}s: %s\n", n, d) }
-  }
+  def smvColMeta(colMeta: (String, String)*): DataFrame =
```
ninjapapa: This method is just for using from python side, so should not be in the `SmvDFHelper`. Just implement in the `SmvPythonProxy`.

src/main/python/smv/schema_meta_ops.py
```diff
@@ -0,0 +1,137 @@
...
+            return [(col.name, getMetaDesc(col.metadata)) for col in self.df.schema.fields]
+        return getMetaDesc(self.getMetaByName(colName))
+
+    def addDesc(self, *colDescs):
+        if not colDescs:
+            raise SmvRuntimeError("must provide (name, description) pair to add")
+
+        addDict = dict(colDescs)
+
+        jdf = self._jPythonHelper.smvColMeta(self.jdf,\
```
laneb: What should happen if the user tries to add a description to a field that doesn't exist? Appears we would ignore it.

src/main/scala/org/tresamigos/smv/SchemaMetaOps.scala
```diff
@@ -16,107 +16,30 @@ package org.tresamigos.smv
...
-      } else df(f.name)
-    }
-
-    df.select(columns: _*)
-  }
-
-  def addDesc(colDescs: Seq[(String, String)]): DataFrame = {
-    require(!colDescs.isEmpty)
-    val colMap = colDescs.toMap
+  def setColMeta(colMeta: Seq[(String, String)]): DataFrame = {
```
ninjapapa: If this method is the only thing left in this class, no need to keep it, just move to `SmvPythonProxy`.

src/main/python/smv/schema_meta_ops.py
```diff
@@ -0,0 +1,137 @@
...
+            [(col.name, json.dumps(setMetaDesc(col.metadata, addDict[col.name])))\
```
laneb: Need to decompose this. Way too much going on for a single statement.

src/main/python/smv/schema_meta_ops.py
```diff
@@ -0,0 +1,137 @@
...
+            for col in self.df.schema.fields if col.name in addDict])
+
+        return DataFrame(jdf, self._sql_ctx)
+
+    def removeDesc(self, *colNames):
+        removeAll = not bool(colNames)
```
laneb: No need to cast to bool.

src/main/python/smv/schema_meta_ops.py
```diff
@@ -0,0 +1,137 @@
...
+        if not removeAll:
+            removeSet = set(colNames)
+
+        jdf = self._jPythonHelper.smvColMeta(self.jdf,\
+            [(col.name, json.dumps(removeMetaDesc(col.metadata)))\
+            for col in self.df.schema.fields if removeAll or col.name in removeSet])
```
laneb: Same point about decomposition.

src/main/python/smv/schema_meta_ops.py
```diff
@@ -0,0 +1,137 @@
...
+    def getLabel(self, colName):
+        if colName is None:
+            return [(col.name, getMetaLabels(col.metadata)) for col in self.df.schema.fields]
+        return getMetaLabels(self.getMetaByName(colName))
+
+    def addLabel(self, colNames, labels):
+        if not labels:
+            raise SmvRuntimeError("must provide a list of labels to add")
+        
+        addToAll = not bool(colNames)
```
laneb: Same point about casting to bool. Also, if `colNames` is empty and thus falsy, we should probably and labels to _no_ columns, not _all_ columns.

src/main/python/smv/schema_meta_ops.py
```diff
@@ -0,0 +1,137 @@
...
+        if not addToAll:
+            addSet = set(colNames)
+
+        jdf = self._jPythonHelper.smvColMeta(self.jdf,\
+            [(col.name, json.dumps(setMetaLabel(col.metadata, labels)))\
+            for col in self.df.schema.fields if addToAll or col.name in addSet])
```
laneb: Would be much clearer to just precalculate what fields you want to add labels to.

src/main/python/smv/schema_meta_ops.py
```diff
@@ -0,0 +1,137 @@
...
+
+        jdf = self._jPythonHelper.smvColMeta(self.jdf,\
+            [(col.name, json.dumps(removeMetaLabel(col.metadata, labels, allLabel)))\
+            for col in self.df.schema.fields if allCol or col.name in removeSet])
+
+        return DataFrame(jdf, self._sql_ctx)
+
+    def colsWithLabel(self, labels = None):
+        def match(meta):
+            return set(labels) <= set(getMetaLabels(meta)) if bool(labels) else not getMetaLabels(meta)
```
laneb: Not obvious what you're trying to express or what `match` means.

src/main/python/smv/schema_meta_ops.py
tianruiwang: Given that their interfaces are already documented in `helpers.py`, do we still need to docstrings here?

src/main/python/smv/schema_meta_ops.py
ninjapapa: Let's error out if can't find field.

tianruiwang: To achieve this, will need:
- option 1: extra time complexity, search whether a given column name exists in the dataframe
- option 2: extra space complexity, store dataframe columns as a dictionary by column names (instead of the current implementation: storing the input queries as a dictionary, which is normally smaller in size)

Which one to choose?

tianruiwang:
Requirements are partially done, remaining things to discuss:
1. whether to document the public methods in `SchemaMetaOps`.
2. how to implement "throw when adding desc/label to a column that doesn't exist".

src/main/python/smv/schema_meta_ops.py
tianruiwang: Or option 3: do it on scala side, where there's already a map structure for dataframe columns.

tianruiwang:
Update: Added checks for column existence. When adding/removing descriptions and labels, any invalid column name will be errored out. As discussed with @ninjapapa, the number of dataframe columns shouldn't be a concern in terms of complexity.

src/main/python/smv/schema_meta_ops.py
laneb: Yeah I think so, particularly since the plan is to remove the Scala implementation. Doesn't need interface documentation like a user-facing API but a quick description of what it is will be helpful in the long term.

laneb: Fine to just c/p from existing documentation.

src/main/python/smv/schema_meta_ops.py
```diff
@@ -88,6 +88,18 @@ def _updateColMeta(self, colShouldUpdate, colUpdateMeta):
...
+        """Check if the given column names exist in the DataFrame
+
+            Will throw if some of the column names are not found
+
+            Args:
+                colNames (list(string)) a list of column names to check
+        """
+        invalidCols = set(colNames) - set(self.df.columns)
+        if invalidCols:
+            raise SmvRuntimeError("column name {} not found".format(", ".join(invalidCols)))
```
laneb: Error will be more helpful if it describes input, like `DataFrame[...] does not have columns ...`

tianruiwang:
Good points! Requirements addressed.

ninjapapa:
@laneb since @tianruiwang may not be able to response to the comments, can we merge this for now?

laneb:
👍

PR #1356, opened 2018-09-24
Remove ascii graph

### LINKED ISSUE #1354: Remove Ascii graph fully

ninjapapa:
Fixed #1354

* Removed ascii-graph dependency
* Moved other SmvGraphUtil tests to python side.


src/test/python/testSmvShell.py
```diff
@@ -30,8 +30,45 @@ def test_shell_df(self):
         self.should_be_same(res, expected)
 
     def test_shell_cmds(self):
-        smvshell.ls()
-        smvshell.lsStage()
+        cmd = smvshell._jvmShellCmd()
+        self.assertEqual(cmd.ls(),
+            """
+stage:
+  (I) stage.modules.CsvStr
```
AliTajeldin: Not sure if this list is deterministic.  It may fail on another system due to the order of discovery on FS.

laneb:
lgtm

src/test/python/testSmvShell.py
ninjapapa: The ordering is deterministic.
The order of the stages is the ordering in the conf file, and the ordering within each stage is sorted by FQN.

PR #1361, opened 2018-09-26
 #1360 fixed cycle dependency bug on python

### LINKED ISSUE #1360: Python module cycle dependency causes stack overflow
The following test case causes stack overflow.
```Python
class CycleA(SmvModule):
    def requiresDS(self):
        return [CycleB]

    def run(self, i):
        return None

class CycleB(SmvModule):
    def requiresDS(self):
        return [CycleA]

    def run(self, i):
        return None
```

The issue is in `DataSetResolver.scala` -> `resolveStack`. Since `resolveStack` is `Seq[SmvDataSet]`, for python modules, they are `SmvExtModulePython`s. Each time `DataSetRepoPython.loadDataSet` is called, a new `SmvExtModulePython` is created on the same python module. Since `SmvExtModulePython` is no more a case class, those `SmvExtModulePython` instances are not the same.

ninjapapa:
Fixed #1360

### DIFF
--- a/src/main/scala/org/tresamigos/smv/DataSetResolver.scala
+++ b/src/main/scala/org/tresamigos/smv/DataSetResolver.scala
@@ -67,18 +67,18 @@ class DataSetResolver(val repos: Seq[DataSetRepo],
    * dependency cycles. Note: we no longer have to worry about corruption of
    * resolve stack because a new stack is created per transaction.
    */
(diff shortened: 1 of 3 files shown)


laneb:
Given that we're using URNs to test uniqueness, will this break down if we have a dependency chain like
`mod:A` -> `link:B` -> `mod:C` -> `mod:B`

ninjapapa:
Good point. Let me create another issue.

Issue #1351, opened 2018-09-21
Remove DataSetRepoScala


ninjapapa:
Child of #1338 

We only need to keep the DataSetRepoPython.

ninjapapa:
The major work here is to move a lot tests from Scala to Python. May need to take multiple PRs to accomplish.

ninjapapa:
* [ ] DataSetMgrTest
* [ ] DQMTest - no need to move, but remove dependency to modules
* [ ] PublishTest
* [ ] SmvAppTest
* [ ] SmvDSPersistTest
* [ ] SmvGraphUtilTest
* [ ] SmvJdbcTest
* [ ] SmvLinkFollowTest
* [ ] SmvLockfileStatusTest
* [ ] SmvMetadataTest
* [ ] SmvModuleLinkHashTest
* [ ] SmvReflectionTest - remove
* [ ] SmvRunInfoCollectorSpec 
* [ ] SmvStagesTest
* [ ] SmvNeedsToRunTest
* [ ] DependencyRuleTest

PR #1363, opened 2018-09-27
I1351 step2 move more tests

ninjapapa:
Step2 of #1351 

- Fixed DQMTest dependency
- Removed Edd persist read back test (never used function, will move into smvMetadata)
- Fixed SmvConfigTest dependency
- Removed ClassCRCTest
- Removed RefectTest data (test already moved to python)
- Moved persist null string test to python
- Moved part of SmvApp tests to python
- Moved all other Scala tests which need to be converted to a separate dir


ninjapapa:
Downstream app itest passed.

Issue #1364, opened 2018-09-27
Drop multi-stage support


ninjapapa:
The multiple stage function, especially the `SmvModuleLink` mechanism, was never adopted by clients. The recent project with multiple stages are simply independent projects under the same repo. 

The reason why this concept never been adopted could be the following:
- Too much for user to learn before using it
- User may want more explicit control on the output data of each "stage", instead of let SMV to manage for them
- Anyhow the multiple stages typically run in multiple steps, multi SmvApps might be more natural than a single SmvApp with multiple stages

We will remove `SmvModuleLink`, publish stage versions, etc. So effectively only support Single-stage project. 

In longer run, we may consider to allow multiple SmvApps under the same project.

yw-yang:
I agree to remove `SmvModuleLink` to simplify the usage. The only concern is `SmvModuleLink` has been widely used in Telstra project metric implementation, is it okay to just remove it or we should inform Telstra side?

ninjapapa:
@yw-yang good to know. 
Since switch between multiple published versions of the up-stream stage when run the down-stream stage is more a function development than production, is Client-T using that function after we delivered? Can we keep `SmvModuleLink` just as a pass through, but remove the config to multiple published version function? The code should still be able to run.

yw-yang:
Exactly, `SmvModuleLink` used in Client-T is just a pass through. I think it's okay to remove the config to multiple published version function and make sure code is still able to run.

yw-yang:
@ninjapapa I just opened a PR to keep `SmvModuleLink` as a pass through. I didn't remove the config to multiple published version function since I found a single module is also supported to be published out, shall we disable this behavior going forward?

ninjapapa:
@yw-yang please see PR comments.

We should still keep publish itself, while removing the read back from published versions, since we do have publish-hive kept. We may remove those command line parameter after #1202

yw-yang:
# Design

The aim of this ticket is to simplify the stage usage. To be backwards compatible, we will only remove the functionality that down-stream module can read back from up-stream stage published versions. This change will not impact any production code that used multi-stage and SmvModuleLink before.

So from user usage perspective, smv will
- still allow multiple stages in conf as `smv.stages=stage1, stage2`
- still allow `-s stage1` command line parameter to specify a stage level run
- still support `SmvModuleLink`, but not support read from published data function of it, so it is just a pass through
- Disable `smv.stages.stage1.version` usage in conf

[...]

Issue #1249, opened 2018-05-25
Create design doc for Simplify flow of DQM validation and SmvDataSet.computeDataFrame


laneb:
The control flow and order of operations in `computeDataFrame` gets pretty confusing due to details of when and whether to compute metadata, force an action to compute DQM results, etc. It's a bit tangled and could use a refactor.

ninjapapa:
Since we removed DQM policy dependency to the result DF of the module, we can remove the "force-action" logic, instead, delay the policy validation to later actions.

ninjapapa:
# High Level Behavior Design

* Metadata will only be calculated and persisted if the DS is NOT Ephemeral
* Make DQM validation part of Metadata calculation. However upper stream non-validated DQM state of ephemeral DS' will also be validated and result persisted in current DS metadata

@laneb @AliTajeldin 
Please review.

AliTajeldin:
As per discussion with @ninjapapa we will "run" the ephemeral DQM rules on a module by module basis, even for ephemeral modules.  That is, if A -> B -> C and B is ephemeral, when A is "run", it will need to evaluate B and that will run the rules for B and the result of B DQM is persisted in B metadata.  If another module also depends on B, it will use the persisted B metadata rather than recompute it.

laneb:
@AliTajeldin Does that differ from what we have today, other than in how the DQM result is persisted?

AliTajeldin:
same.  We just no longer need the force flag.

ninjapapa:
With removing the force run, current module's rdd calculation will not be able to trigger its DQM's policy check. So need to maintain a queue on App level to keep tracking the calculated ephemeral modules which have not been policy checked, and check the queue when a persisted module get calculated.

ninjapapa:
For this issue, yes. No DQM policy checks if all modules are ephemeral. However, we should re-define persist of a output module (not current SmvOutput, but dedicated output module) as write to output DB or channels, which should never be ephemeral.

jacobdr:
I think this would have broken our last client use case, so maybe we should discuss in person.
In that case we wanted to re-run ephemeral inputs hourly to detect early DQM issues, but defer full graph execution for the previous day to during a slow hour on the cluster (usually overnight).
Maybe since the runtime params were different for each invocation this would cache bust, but not sure how that plays in to the refactor.

ninjapapa:
[...]

PR #1366, opened 2018-09-28
I1351 step3 move all tests

ninjapapa:
Move all scala tests which used SmvModule to python, except

- SmvMetadataTest - will recreate on python side after #1249 
- SmvStagesTest - decide when #1364 has a conclusion


src/test/python/testRunCmdLine.py
```diff
@@ -12,34 +12,92 @@
 # limitations under the License.
 
 from test_support.smvbasetest import SmvBaseTest
+from py4j.protocol import Py4JJavaError
 
-
-class RunModuleFromCmdLineTest(SmvBaseTest):
-    modUrn = 'runmod.stage1.modules.A'
+class RunCmtLineBaseTest(SmvBaseTest):
```
AliTajeldin: `Cmt` --> `Cmd`

Issue #1365, opened 2018-09-27
Design of `SmvResult` class to allow generic output of modules


AliTajeldin:
This ticket to formalize the design of the `SmvResult` class to allow us to create generic `SmvModules` (e.g. ones that produce Pandas, H2O dataframes, models, etc).

ninjapapa:
@AliTajeldin done?

PR #1367, opened 2018-09-28
#1365: added smv result class/docs

AliTajeldin:
Design for `SmvResult` class.  The plantuml in PR is shown below as well:

class_diagram

There are a couple of possible variations/TBD:
* Make current `SmvModule` the base class instead of adding `SmvGenericModule`.
* add a `canPersist()` method to `SmvResult` that would notify SMV that these types of results can not be persisted so therefore can not have `isEphermeral()` set to false.
* move the `persist`/`unpersist` into the module itself (still add the methods though).  Less classes to maintain but harder to reuse persistence engines (e.g. we can just have a generic `PickleResult` mix-in that all serializable types can use directly).  Then again that can be mixed in into the `*Module` ass as well

### DIFF
--- a/docs/dev/DataSetMgr/DataSetMgr_class.puml
+++ b/docs/dev/DataSetMgr/DataSetMgr_class.puml
@@ -54,8 +54,8 @@ package Scala {
       run(i: runParams): DataFrame
       resolvedRequiresDS: Seq[SmvDataSet]
     }
-    class SmvExtModulePython -|> SmvModule
-    class SmvModule -L|> SmvDataSet
+    SmvExtModulePython -|> SmvModule
+    SmvModule -L|> SmvDataSet
   }
   class SmvExtModule extends SmvModule
   class SmvModuleLink extends SmvModule
--- a/docs/dev/SmvResult/SmvResult.md
+++ b/docs/dev/SmvResult/SmvResult.md
@@ -2,10 +2,41 @@
 
 There are circumstances in which it would be useful for an `SmvModule` to return a result which is not a `DataFrame`. For example, when using SMV for modeling it would be helpful to train the model in one module and score data with the model in another module downstream. In this scenario, we would also like to persist the model as an intermediate result as we would with a `DataFrame`.
 
-## Pickle & persist
+We currently handle the "model" special case by converting the model result into a string (by pickle) and stuffing that into a single row `DataFrame`.
+This approach will not scale to other types well.  For example, we would like to support "generic" modules that may depend-on and return types such
+as Pandas data, H2O data, none-serializable models, etc.
 
-If possible, we would like to use the same mechanisms for persisting these generic results as we do for persisting `DataFrames`. We can accomplish this by serializing the user's result and saving the serialization as a single row `DataFrame`. Since this feature is only needed for Python `SmvModules`, we can use `pickle` for the serialization of arbitrary results. Of course, this requires that the result is safely picklable.
+# Solution
+The solution is to wrap all module results into a concrete implementation of `SmvResult` class.
+`SmvResult`s would be responsible for persisting/reading result data.  The `SmvModule` will no longer assume results/inputs are Spark `DataFrame`.
 
-## SmvResultModule
+Each concrete `SmvModule` type would have to specify the corresponding `SmvResult` class that can handle the result of it's `run` method.
+The current `SmvModule` can continue to be specific to Spark modules and can be bound to `SmvDataframeResult` result.
+Other modules types can specify different result types.
+For example, `SmvPandasModule` can specify `SmvPandasResult` as it's result type.
+Downstream modules that depend on a instance of `SmvPandasModule` can assume that they will receive a pandas `DataFrame` as input.
 
-When the result is read back as an input, we need some indication that the `DataFrame` contains a pickled object. Modules that return generic results will inherit from `SmvResultModule` (name will probably be updated upon further consideration).
+## Example client code
+```python
+class M1_S(SmvModule):
+   """This is a spark dataframe module"""
+   def run(i):
+     return DataFrame(...) # spark dataframe
+
+class M2_P(SmvPandasModule):
+   """This is a pandas dataframe module"""
+   def run(i):
+     return pandas.createDataFrame(...) # pandas dataframe
+
+class M3_J(SmvModule):
+   """spark module that combines pandas+spark dataframes"""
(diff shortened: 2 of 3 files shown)


ninjapapa:
If we implement this in python, why we need `SmvModule` and `SmvPandasModule` to be two different classes?

laneb:
IMO we should call it `SmvPersistenceStrategy`, not `SmvResult`. Could have multiple persistence strategies for the same result type, e.g. parquet or orc for `DataFrame`. Otherwise lgtm.

AliTajeldin:
* `SmvModule` and `SmvPandasModule` need to be different because they provide a different return type of `resultType()` method.  `SmvPandasModule` returns `SmvPandasResults` which knows how to persist/read pandas data.  Same pattern would apply to other types of `Smv*Module`
* As for calling this a "persistence strategy", I think the persistence strategy is orthogonal to the result type.  For example, the result type can be used in a UI to inform the user the "type" of the incoming data (e.g. spark dataframe, model, etc).  That is different than the persistence strategy used (e.g. csv file, parquet, etc).  For the short term, there is only one persistence strategy per result type, but others could be added in the future.

laneb:
Should probably make the result type and the persistence strategy orthogonal at this crossroad then, instead of migrating client code down the line. But does it make sense for SMV to even worry about the result type? If the result type is purely a UI consideration, maybe it should be pushed into the client application.

ninjapapa:
@AliTajeldin are we concluded that the `SmvResult` should NOT be a class separated from `SmvGenericModule`?

AliTajeldin:
I think the last discussion was to use some form of IO strategy class (which would actually match the hierarchy of the `SmvResult` above).
I'll update the design.

ninjapapa:
I got a new idea when I was trying to decompose current `SmvDataSet`. 

As current `SmvDataSet` at lease handles the following thing:
* Dependency and resolve dependency (tight to the transactions)
* Run module (rdd)
* Data persist/unpersist
* Data publish
* MetaData creation (will move both DQM and EDD in MetaData)
* MetaData persist/unpersist (including history)

Only the first item provides the user interface methods. We will for sure separating out the data persisting stuff to the `IOStratagy", but the MetaData stuff (or can consider as run-info) is closely related to "Run module". Ideally we should separate the MetaData part out. The natural way to do that is actually create a `ModuleRunner` object to provide the Run method and Metadata management.

How about:
* `SmvGenericModule` only covers the user interface methods, and dependency resolving
* `IOStratagy` handle the data persist/unpersist
* `ModuleRunner` as a util Object with a method `run(m:SmvGenericModule, stgy:IOStratagy)`

ninjapapa:
Never mind. Doesn't work, since each run actually need to run the whole chain, and we may have different IO strategies along the chain.

AliTajeldin:
making some updates, **not** ready for review yet

AliTajeldin:
@ninjapapa @laneb ready for re-review.

docs/dev/SmvIoStrategy/SmvIoStrategy.md
```diff
@@ -0,0 +1,113 @@
...
+Note: if a module is not ephemeral, it **must** define an `persistStrategy` method that returns a valid strategy.  Ephemeral modules can ignore that method.
+
+## `SmvIoStrategy` implementation
+Since the concrete dataset class needs to provide an instance of `SmvIoStrategy`, we can use the constructor of the strategy to capture the variant
+in the required arguments to the strategy.  For exmaple, HDFS strategy may require a directory whereas a hive strategy may require a schema and
+table name.  These areguments can be supplied to the constructor as the caller (concrete dataset) will know what is required and will have the info.  This has two advantages:
+* Do not have to make a kitchen sink interface on dataset to support all possible information needed by strategies.
+* Can easily extend the types of datasets and strategies in the future without changing core SMV
+
+`SmvIoStrategy` will also provide interfaces for read/write of both the raw output and metadata.
```
ninjapapa: I don't think data read/write and metadata read/write should be in the same `SmvIoStrategy`. Since it should be perfect fine to store the data in hive while leave metadata on hdfs dirs. Actually metadata itself can be considered as another data, and has its own `SmvIoStrategy`.

docs/dev/SmvIoStrategy/SmvIoStrategy.md
```diff
@@ -0,0 +1,113 @@
...
+
+From the io strategy point of view, there is no difference between intermediate data persist and publishing.
```
ninjapapa: Who own the logic of deriving the file path and names? The caller or the `SmvIoStraytegy`?

docs/dev/SmvIoStrategy/SmvIoStrategy.md
AliTajeldin: good idea.  Will add a `metadataPersistStrategy` method and make that a flag in the constructor to tell that strategy if it is persisting the metadata or raw data (for ones with such support).

docs/dev/SmvIoStrategy/SmvIoStrategy.md
AliTajeldin: Ultimately, it would be the io strategy with some info coming from the module/app (e.g. data dir)

AliTajeldin:
Updated the design doc to separate out the raw data vs metadata persist strategies.

ninjapapa:
Since metadata persist with history, we could actually generalize that idea and allow `SmvIoStrategy` to support `persistWithHistory` and `persistWithoutHistory`, either through a flag or other ways to switch.

Another thing need to consider, not related to `SmvIoStrategy` but does related to the `Result` type. For example, the delay-action nature of Spark Dataframe, and the inline Dqm itself. Those things are result-type specific, but unrelated to the detail IO strategy of how to persist. Should we keep create another level of ABC, say, `SmvGenericSparkDfModule`, which has the delay validation logic and Dqm?

AliTajeldin:
Yah, I agree, the whole current `SmvDataset` tree and below is really `SmvSparkDataset` which is a child of `SmvGenericNode/Module`.  Somehow the logic for delay/dqm/etc needs to be captured in the `SmvSparkDataset` root or below rather than in the resolver or app.  Need to think about this a bit.

AliTajeldin:
@ninjapapa as per discussion today, merged this under `docs/SmvGenericModule`.  That will be the catch all for the python generic module interface as well.

PR #1369, opened 2018-09-28
I1364 delete publish stage

### LINKED ISSUE #1346: Remove Scala interface from docs
Child of #1338

### LINKED ISSUE #1364: Drop multi-stage support

yw-yang:
Fixes #1364 

All tests passed.
 I only remove stage version part in this PR. Since we can also publish module output besides stage output, I didn't get rid of public version function. @ninjapapa let me know if any suggestion.


src/main/scala/org/tresamigos/smv/SmvDataSet.scala
```diff
@@ -663,13 +663,11 @@ abstract class SmvDataSet {
...
-  private[smv] def stageVersion()                   = parentStage flatMap { app.smvConfig.stageVersions.get(_) }
-
   /**
-   * Read the published data of this module if the parent stage has specified a version.
-   * @return Some(DataFrame) if the stage has a version specified, None otherwise.
+   * Read the published data of this module
+   * @return Some(DataFrame) if the module has a version specified, None otherwise.
    */
-  private[smv] def readPublishedData(version: Option[String] = stageVersion): Option[DataFrame] = {
+  private[smv] def readPublishedData(version: Option[String]): Option[DataFrame] = {
```
ninjapapa: Is this method still needed?

src/test/python/testSmvFramework.py
```diff
@@ -248,34 +248,3 @@ def test_ephemeral_module_depends_on_need_to_run_also_need(self):
         self.df(fqn, True)
         self.deleteModuleOutput(self.load(fqn0)[0]) # deleting persist files made M1 need to run
         self.assertTrue(self.load(fqn)[0].needsToRun())
-
```
ninjapapa: This test should still kept. We just removed read versioned published data function, should still allow publish itself.

src/main/scala/org/tresamigos/smv/SmvDataSet.scala
yw-yang: This method is also used for reading the data published by an output module, see
https://github.com/TresAmigosSD/SMV/blob/master/src/main/scala/org/tresamigos/smv/SmvApp.scala#L352
Maybe we can just keep it

yw-yang:
Close now, will put a design in the issue first

yw-yang:
Comments have been addressed, please review again

src/main/scala/org/tresamigos/smv/DataSetResolver.scala
```diff
@@ -101,9 +101,6 @@ class DataSetResolver(val repos: Seq[DataSetRepo],
             if (dsStage == l.smvModule.parentStage.get)
               throw new SmvRuntimeException(msg.sameStageLink(ds.urn, dsStage))
           case _ =>
-            val depStage = dep.parentStage.get
```
ninjapapa: If we only handle one match case, instead of use `foreach`, we can use `collect` as following:
```scala
ds.resolvedRequiresDS collect {i: SmvModuleLink =>
          if (dsStage == l.smvModule.parentStage.get)
              throw new SmvRuntimeException(msg.sameStageLink(ds.urn, dsStage))
        }
```

yw-yang: Changes made and all tests passed

Issue #1370, opened 2018-09-29
Move SmvApp functions to python side


ninjapapa:
Since python side has the access to all the DS instance already, it should be pretty easy to implement the SmvApp functions in python side also. Will create the python side functions one by one, and remove the scala version when ready.

PR #1373, opened 2018-10-01
I1370 step2 reduce smvapp to run only

ninjapapa:
This is step 2 of #1370 

Reduced SmvApp on scala side to only do:

* Hold singletons of log, dsm and smvConfig
* The run method

Next step will move the run method to python side also.

### DIFF
--- a/src/main/scala/org/tresamigos/smv/CreateDF.scala
+++ b/src/main/scala/org/tresamigos/smv/CreateDF.scala
@@ -1,0 +1,39 @@
+/*
+ * This file is licensed under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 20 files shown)


src/main/scala/org/tresamigos/smv/CreateDF.scala
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.tresamigos.smv
+
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.tresamigos.smv.dqm.{ParserLogger, TerminateParserLogger}
+
+private[smv] object CreateDF {
```
AliTajeldin: * class names should be nouns (e.g. `DfCreator` or `DfFactory`).
* Missing doc to describe what/why of this singleton class.

PR #1379, opened 2018-10-02
Removed SmvModuleLink from Scala

### LINKED ISSUE #1377: Make SmvModuleLink a pure pass through
In PR #1369, read back versioned data through the SmvModuleLink mechanism we are removed. So other than links still show in the list or graph when print/plot the modules, they are just pass the target module down. 

Since all the client project I know use the SmvModuleLink in the following way:
```python
M2 = SmvModuleLink(M1)
```

We can just replay the `SmvModuleLink` class by an identity function. 

There will be no behavior change of any projects already delivered, except when use try to list out modules in shell. 

@jacobdr @yw-yang @laneb could you confirm the client-T code only use `SmvModuleLink` in above way?

ninjapapa:
Fixed #1377

Made `SmvModuleLink` in python as simply an identity function, and removed the entire class from Scala side.


src/main/python/smv/smvdataset.py
```diff
@@ -672,59 +666,8 @@ def run(self, i):
...
-        SmvExtModuleLink(fqn) is equivalent to SmvModuleLink(SmvExtDataSet(fqn))
-
-        Args:
-            fqn (str): fqn of the the Scala SmvDataSet
-
-        Returns:
-            (SmvModuleLink): link to the Scala SmvDataSet
-    """
-    return SmvModuleLink(SmvExtDataSet(refname))
+def SmvModuleLink(target):
```
AliTajeldin: 🎉

PR #1380, opened 2018-10-02
I1374 rm r

### LINKED ISSUE #1374: Remove R support
The Smv R support were used in a single client PoC project to demo the possibilities. Given the fact that Spark itself has python side grow much faster than R side, and we have no plan to make Smv R support catching up with the Python side, we will drop the SMV R support totally.

ninjapapa:
Fixed #1374 

Removed R support.


ninjapapa:
Close to delete some code related to R.

PR #1381, opened 2018-10-02
I1370 last step - moved run to python

### LINKED ISSUE #1370: Move SmvApp functions to python side

ninjapapa:
Fixed #1370 

This is the last step of #1370.

Since we will not change behaviors of this issue, we still leave some of the functions for handling command line on Scala side. Since we moved the run method itself to Python, we can than move those functions one by one with the behavior changes needed for them as in separated issues.

### DIFF
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -452,6 +452,12 @@ class SmvApp(object):
         """Returns a Scala None value"""
         return self.scalaOption(None)
 
+    def _scala_seq_to_list(self, j_seq):
+        """Convert Scala Seq to Python list
+        """
+        j_list = self._jvm.scala.collection.JavaConversions.seqAsJavaList(j_seq)
+        return [x for x in j_list]
+
     def createDFWithLogger(self, schema, data, readerLogger):
         return DataFrame(self._jvm.DfCreator.createDFWithLogger(
             self.sparkSession._jsparkSession,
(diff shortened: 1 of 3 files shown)


src/main/python/smv/smvapp.py
```diff
@@ -452,6 +452,12 @@ def scalaNone(self):
         """Returns a Scala None value"""
         return self.scalaOption(None)
 
+    def _scala_seq_to_list(self, j_seq):
```
AliTajeldin: Does this mean we can remove a bunch of functions from the proxy that just do `seq -> list` conversions?

src/main/python/smv/smvapp.py
```diff
@@ -498,5 +504,55 @@ def remove_source(self, project_path):
...
+        """Execute as dry-run if the dry-run flag is specified.
+            This will show which modules are not yet persisted that need to run, without
+            actually running the modules.
+        """
+        
+        if(self._cmd_line().dryRun()):
+            # Find all ancestors inclusive,
+            # filter the modules that are not yet persisted and not ephemeral.
+            # this yields all the modules that will need to be run with the given command
+            mods_with_ancestors = self._scala_seq_to_list(self.j_smvApp.modulesToRunWithAncestors())
```
AliTajeldin: neat re-use of this function.

ninjapapa:
Merged too quick. Test failed :-( 
Create another issue to fix.

PR #1395, opened 2018-10-09
I1389 run conf hash

### LINKED ISSUE #1389: Fix _smvGetRunConfigHash in SmvDataSet
Search `_smvGetRunConfigHash` in code will find that it was used in `SmvDataSet` without definition. Since it was in try-catch logic, the hash logic to handle `SmvRunConfig` mix-in case was simply ignored.

Instead of implementing that method, here is a better way to address it:
* Define a concrete `requiresConfig` with all the possible run conf keys in `SmvRunConfig` mix in
* Remove all the other methods in `SmvRunConfig`
* Remove `_smvGetRunConfigHash` part of the hash calculation totally, since `config_hash` part already covers all the possible run configs which could be used
* Document in user doc that `SmvRunConfig` mix-in will be deprecated, and it is simply added a requires-all `requiresConfig` method, which may cause unnecessary module recalculations

ninjapapa:
Fixed #1389 

Instead of use `SmvRunConfig` mix-in to override `requiresConfig`, use a "private" function `_all_run_conf_keys`, since client code typically has `MyModule(SmvModule, SmvRunConfig)` form, which `SmvModule` overrides `SmvRunConfig`.


src/main/python/smv/runconfig.py
```diff
@@ -29,25 +27,6 @@ class SmvRunConfig(object):
...
-            return None
-        sval = runConfig.strip().lower()
-        return (sval == "1" or sval == "true")
-
-    def _smvGetRunConfigHash(self):
-        """return the app level hash of the all the current user config values"""
-        app = SmvApp.getInstance()
-        app.warn("SmvRunConfig is deprecated. Accessors like smvGetRunConfig are now available directly on SmvDataSet.")
-        return app.j_smvPyClient.getRunConfigHash()
+    def _all_run_conf_keys(self):
```
AliTajeldin: `SmvRunConfig` is deprecated.  We should either preserve it's interface for existing projects or remove it all together.

ninjapapa:
close to fix test.

ninjapapa:
I forced pushed the branch, so no way to reopen this. Will start a new PR and point back.

Issue #1392, opened 2018-10-09
Fail to run SMV Example App 


yw-yang:
After I installed latest smv and tried to run the example app
`smv-init -s MyApp`
`smv-run --run-app`

Below error occurs. I can trigger other project successfully, is the example app out-dated? Since it is noted in the SMV homepage, could we have a fix on it?
```
Modules to run/publish
----------------------
stage1.employment.EmploymentByState
----------------------
Traceback (most recent call last):
  File "/Users/Documents/Programming/smv/SMV_package/SMV_master/tools/../src/main/python/scripts/runapp.py", line 17, in <module>
    SmvDriver().run()
  File "/Users/Documents/Programming/smv/SMV_package/SMV_master/src/main/python/smv/smvdriver.py", line 54, in run
    self.main(app, driver_args)
  File "/Users/Documents/Programming/smv/SMV_package/SMV_master/src/main/python/smv/smvdriver.py", line 35, in main
    app.run()
  File "/Users/Documents/Programming/smv/SMV_package/SMV_master/src/main/python/smv/smvapp.py", line 629, in run
    or self._generate_output_modules(mods, collector)
  File "/Users/Documents/Programming/smv/SMV_package/SMV_master/src/main/python/smv/smvapp.py", line 604, in _generate_output_modules
    self._module_rdd(m, collector)
  File "/Users/Documents/Programming/smv/SMV_package/SMV_master/src/main/python/smv/smvapp.py", line 564, in _module_rdd
    False # quick run
  File "/Users/Documents/Programming/smv/spark-2.3.1-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/java_gateway.py", line 1257, in __call__
  File "/Users/Documents/Programming/smv/spark-2.3.1-bin-hadoop2.7/python/lib/pyspark.zip/pyspark/sql/utils.py", line 63, in deco
  File "/Users/Documents/Programming/smv/spark-2.3.1-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/protocol.py", line 328, in get_return_value
py4j.protocol.Py4JJavaError: An error occurred while calling o151.rdd.
: org.tresamigos.smv.SmvRuntimeException: There was an error executing Python code
 at org.tresamigos.smv.python.InterfacesWithPy4J$class.getPy4JResult(Py4JInterface.scala:28)
[...]

PR #1396, opened 2018-10-10
 #1392 fixed the --run-app bug

ninjapapa:
The fix is simple, but the root cause of this bug is very deep, which relates on the way python modules get dynamically loaded. 

The problem was that when `create_graph_dot` get called out side of the if condition, it ran even if we just do `--run-app`. When it is called, one DSM call was made to load all modules, while the real `generate_output` need to load the output modules the second time. When the output module get reloaded, the dependency didn't, since they were loaded once already in the `create_graph_dot` call.

### DIFF
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -566,8 +566,8 @@ class SmvApp(object):
     def _generate_dot_graph(self):
         """Genrate app level graphviz dot file
         """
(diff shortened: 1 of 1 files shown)


ninjapapa:
@laneb this PR fixed the bug, but could you review the reasoning part? I'm not sure whether I really understand the way python modules get loaded.

ninjapapa:
Fixed this specific issue #1392 here.
Opened another issue #1398 for further investigation.

PR #1399, opened 2018-10-11
I1389 run conf hash

### LINKED ISSUE #1389: Fix _smvGetRunConfigHash in SmvDataSet

ninjapapa:
This replaced PR #1395 
Fixed #1389 

Since I rebased the branch to master on my local repo, have to do a forced push, and restart a PR. The new stuff to be reviewed is in 3939c93


AliTajeldin:
@ninjapapa ready to go, just needs a merge.

src/test/python/testDynamicRunConfig.py
```diff
@@ -49,7 +50,17 @@ def test_run_module_by_name_with_run_config(self):
         expected = self.createDF('src:String', 'cmd')
         self.should_be_same(expected, df)
 
+<<<<<<< HEAD
```
laneb: Looks like one of the conflicts wasn't resolved.

ninjapapa: Forgot to save from code, before git add....

PR #1403, opened 2018-10-11
I1346 rm scala doc

### LINKED ISSUE #1346: Remove Scala interface from docs

ninjapapa:
Fixed #1346


AliTajeldin:
I think this ticket is only for removing reference to scala docs from user guide?
We should also remove the reference to the scala docs from the `gh_pages` branch and all references to publishing scaladocs (we do that in a couple of scripts).  Just search for `scaladocs` under SMV.  But I think that should be a separate ticket.

Issue #1347, opened 2018-09-20
Simplify Scala SmvDataSet to minimal


ninjapapa:
Child of #1338 

This issue is to further prepare for removing Scala side DS management. 

Will take multiple steps:
* Remove (or convert) scala unittests for DS framework
* Remove unnecessary attributes and methods from scala `SmvDataSet`, `SmvModule` classes
* Cleanup

ninjapapa:
Need to put this on hold, since the cost of converting the framework to Python is too high without simplify DQM and persisting mechanism.

Depends on:
* #1259
* #1249 
* #924

ninjapapa:
Now we are ready to take on this.

Steps:
* [ ] Move csv persist handling out to a `PersistStgy` class
* [ ] Remove DQM valid
* [ ] Move meta handling to a `SmvMetaHandler` class
* [ ] Mode edd process to meta data

PR #1414, opened 2018-10-15
I1347 step1 persist stgy

ninjapapa:
Step 1 out of 4 of #1347

* Introduced `DataPersistStgy` and an implementation `DfCsvOnHdfsStgy`
* Used `DfCsvOnHdfsStgy` in `SmvDataSet`
* Removed Scala side `SmvModule`


ninjapapa:
As this issue is to simplify the `SmvDataSet` class so that it can be moved, so yes, `DataPersistStgy` is just a scaffolding. The real IoStrategy will be implemented on python side after the design is done.

PR #1419, opened 2018-10-17
I1417 python resolver

### LINKED ISSUE #1417: Create Python side DataSetResolver
Child of #1338

Need to create a barebones SmvDataSet2 in pure Python and a Python version DataSetResolver. This issue will not replace any Scala side framework, instead simply to make a python version module resolve mechanism working.

ninjapapa:
Fixed #1417 

This issue and PR is just to introduce the python side resolver. Without making Python side `SmvDataSet` runnable, we can't switch on the function introduced by this PR yet. So no behavior change on the code at all for now.

### DIFF
--- a/src/main/python/smv/datasetresolver.py
+++ b/src/main/python/smv/datasetresolver.py
@@ -1,0 +1,47 @@
+#
+# This file is licensed under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


src/main/python/smv/datasetresolver.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from datetime import datetime
+
+class DataSetResolver:
```
AliTajeldin: needs a doc but can be done in next PR as this is not enabled at the moment.

PR #1420, opened 2018-10-17
I1347 step3 edd to meta

ninjapapa:
Step 3 of #1347 

We use to persist 
* Data (csv, schema)
* DQM (.valid)
* EDD (.edd) optional
* Meta (.meta, .history)

Step 2 removed `.valid`. This step removed `.edd` and moved edd result to another element of meta. 

Another change:
User meta is now **an element** of `SmvMetaData`, instead of is the `SmvMetaData`.

### DIFF
--- a/src/main/scala/org/tresamigos/smv/SmvDataSet.scala
+++ b/src/main/scala/org/tresamigos/smv/SmvDataSet.scala
@@ -116,9 +116,6 @@ abstract class SmvDataSet {
   /** full name of hive output table if this module is published to hive. */
   def tableName: String = throw new IllegalStateException("tableName not specified for ${fqn}")
 
-  /** Objects defined in Spark Shell has class name start with $ **/
-  private val isObjectInShell: Boolean = this.getClass.getName matches """\$.*"""
-
   /** Hash computed from the dataset, could be overridden to include things other than CRC */
   def datasetHash(): Int = {
     val _instanceValHash = instanceValHash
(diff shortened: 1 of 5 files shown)


src/main/scala/org/tresamigos/smv/SmvDataSet.scala
```diff
@@ -401,13 +363,19 @@ abstract class SmvDataSet {
...
+    dfOpt:       Option[DataFrame],
+    validResOpt: Option[DqmValidationResult],
+    genEdd: Boolean
+  ): SmvMetadata = {
+    val resMetadata = new SmvMetadata()
+    dfOpt match {
+      // if df provided, get user metadata and if genEdd, add Edd result
+      case Some(df) => {
+        resMetadata.addUserMeta(getOrCreateUserMetadata(df))
+        if (genEdd) resMetadata.addEddResult(df.edd.toMetaArray)
```
AliTajeldin: do we need to cache the edd result the same way we did for user meta?  I think there was a reason we did that for the user meta that I do not recall.

ninjapapa: User meta may trigger actions on the DF. All other meta do not. In the sense EDD is similar to user meta. I longer run we could actually make the EDD command line flag to add a user meta to all the modules. Or actually may remove the EDD command line flag all together.

ninjapapa: The comments of `userMetadataCache`:
```
Cache metadata so we can reuse it between validation and persistence 
```

However currently both validation and persistence are in `runDqmAndMeta` method, and user meta calculation is just did once in that method. 

The real place the cache makes difference is on ephemeral modules, whose `runDqmAndMeta ` might be called multiple times. 

Instead of manage the cache in `SmvDataSet`, since `runDqmAndMeta` is the only place we persist meta data to file, we should always try to read-back from persisted file first. 

Will make the change.

Issue #1421, opened 2018-10-17
Move entire SmvDataSet to Python side


ninjapapa:
* [x] Create `runModule2` and basic `SmvMoudleRunner` with handling basic module run
* [x] Add persisting and meta
* [x] Implement other current `SmvDataSet` methods
* [ ] Shift to `runModule2`
* [ ] Remove scala `SmvDataSet` and cleanup

PR #1424, opened 2018-10-21
I1421 step1 basic run2

ninjapapa:
Step 1 of #1421 

* Created `ModulesVisitor`, `SmvModuleRunner`
* Implement basic module run with df cache and dqm validation
* NO persist and runinfo yet

### DIFF
--- a/src/main/python/smv/smvdataset.py
+++ b/src/main/python/smv/smvdataset.py
@@ -22,6 +22,8 @@ import sys
 import traceback
 import binascii
 import json
+import queue
+
 from collections import OrderedDict
 
 from smv.dqm import SmvDQM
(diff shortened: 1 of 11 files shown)


src/main/python/smv/smvdataset.py
```diff
@@ -75,6 +88,36 @@ def tableName(self):
...
+
+    def _build_queue(self, roots):
+        """Create a depth first queue with order for multiple roots"""
+        _queue = OrderedDict()
+        def _add_to(mod, q):
+            if (len(mod.resolvedRequiresDS) > 0):
+                for m in mod.resolvedRequiresDS:
+                    _add_to(m, q)
+            q.update({mod: True})
+        for m in roots:
```
AliTajeldin: since `_add_to` is recursive, wouldn't two roots of same level cause the queue not to be DFS?

ninjapapa: Actually it handles that case well. Since the recursive logic guarantees that modules are always in front of the ones depend on them, this algorithm is not exactly DFS as in 
https://en.wikipedia.org/wiki/Depth-first_search

But it is what we need.

ninjapapa: As discussed, will convert to an non-recursive version

PR #1438, opened 2018-10-26
I1427 edd prop

### LINKED ISSUE #1427: Make edd flag a prop instead of a command line flag
Currently we have the command line flag `--edd` to trigger all modules to calculate edd and persist as run. 

This issue will change it to a config prop, `smv.forceEdd = True`. By doing that we will be able to set it dynamically for each run in interactive mode.

#1421 will remove the support to `--edd`. This issue will add the config prop.

ninjapapa:
Fixed #1427

### DIFF
--- a/src/main/python/smv/smvconfig.py
+++ b/src/main/python/smv/smvconfig.py
@@ -143,6 +143,13 @@ class SmvConfig(object):
     def stage_names(self):
         return self._split_prop("smv.stages")
 
+    def force_edd(self):
+        flag = self.merged_props().get("smv.forceEdd")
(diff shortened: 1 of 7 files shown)


src/main/python/smv/smvconfig.py
```diff
@@ -143,6 +143,13 @@ def user_libs(self):
     def stage_names(self):
         return self._split_prop("smv.stages")
 
+    def force_edd(self):
```
AliTajeldin: should probably have a `get_prop_as_bool` method defined here and this method just becomes `return get_prop_as_bool("smv.forceEdd")`

PR #1444, opened 2018-10-27
I1441 more scala clean

### LINKED ISSUE #1441: Further Scala cleanup
The following things can be removed from Scala side

* SmvConfig
* SmvReportIO
* IAnyInputStream

ninjapapa:
Fixed #1441

* Removed Scala side mvConfig and IAnyInputStream interface.
* Removed most of Scala side write to file helpers (since not been called by python side, so save to remove)
* Still kept SmvReportIO, since smvExportCsv still uses it


ninjapapa:
Found that we should be able to remove SmvApp totally. Close to do it.

ninjapapa:
Reopen. 
Simplified scala SmvApp be a simple holder on SC and log.

PR #1452, opened 2018-10-29
I924 parquet io

### LINKED ISSUE #924: Create SmvParquetIoStrategy to persist intermediate results

ninjapapa:
Fixed #924 

Created `SmvParquetOnHdfsIoStrategy`. Will create another issue to switch the default persist to parquet.

### DIFF
--- a/src/main/python/smv/smviostrategy.py
+++ b/src/main/python/smv/smviostrategy.py
@@ -38,20 +38,18 @@ class SmvIoStrategy(ABC):
     def remove(self):
         """Remove persisted file(s)"""
 
-# TODO: add lock, add publish
-class SmvCsvOnHdfsIoStrategy(SmvIoStrategy):
-    def __init__(self, smvApp, fqn, ver_hex, csv_path=None):
-        self.smvApp = smvApp
-        self.fqn = fqn
-        self.versioned_fqn = "{}_{}".format(fqn, ver_hex)
-        if (csv_path is None):
-            output_dir = self.smvApp.all_data_dirs().outputDir
-            self._csv_path = "{}/{}.csv".format(output_dir, self.versioned_fqn)
-        else:
-            self._csv_path = csv_path
 
-        self._schema_path = re.sub("\.csv$", ".schema", self._csv_path)
-        self._lock_path = self._csv_path + ".lock"
+class SmvFileOnHdfsIoStrategy(SmvIoStrategy):
+    def __init__(self, smvApp, fqn, ver_hex, postfix, file_path=None):
+        self.smvApp = smvApp
+        if (file_path is None):
+            versioned_fqn = "{}_{}".format(fqn, ver_hex)
+            output_dir = self.smvApp.all_data_dirs().outputDir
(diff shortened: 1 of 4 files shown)


src/main/python/smv/smviostrategy.py
```diff
@@ -38,20 +38,18 @@ def isPersisted(self):
     def remove(self):
         """Remove persisted file(s)"""
 
-# TODO: add lock, add publish
-class SmvCsvOnHdfsIoStrategy(SmvIoStrategy):
-    def __init__(self, smvApp, fqn, ver_hex, csv_path=None):
+
+class SmvFileOnHdfsIoStrategy(SmvIoStrategy):
```
AliTajeldin: need a doc do describe this base class and what is the derived classes supposed to overwrite/provide.

src/main/python/smv/smviostrategy.py
```diff
@@ -60,32 +58,53 @@ def _smvLock(self):
...
+    def remove(self):
+        self.smvApp._jvm.SmvHDFS.deleteFile(self._file_path)
+
+
+class SmvCsvOnHdfsIoStrategy(SmvFileOnHdfsIoStrategy):
+    def __init__(self, smvApp, fqn, ver_hex, file_path=None):
+        super(SmvCsvOnHdfsIoStrategy, self).__init__(smvApp, fqn, ver_hex, 'csv', file_path)
+
+    @property
+    def _schema_path(self):
```
AliTajeldin: Where is this file being written?  I can see it being deleted, but not written.

src/main/python/smv/smviostrategy.py
```diff
@@ -94,22 +113,27 @@ def isPersisted(self):
...
-        return self._jvm.SmvHDFS.exists(self.path)
 
-    def remove(self):
-        self._jvm.SmvHDFS.deleteFile(self.path)
\ No newline at end of file
+class SmvParquetOnHdfsIoStrategy(SmvFileOnHdfsIoStrategy):
+    def __init__(self, smvApp, fqn, ver_hex, file_path=None):
+        super(SmvParquetOnHdfsIoStrategy, self).__init__(smvApp, fqn, ver_hex, 'parquet', file_path)
+
+    def read(self):
```
AliTajeldin: There seems to be a mismatch between the read/write names.  We should be consistent with the names of the functions that the derived classes would overwrite.  Either both start with `_` or both without.

src/main/python/smv/smviostrategy.py
ninjapapa: will do.

src/main/python/smv/smviostrategy.py
ninjapapa: Scala side call below wrote 2 files, csv and the schema file:
`self.smvApp.j_smvPyClient.persistDF(self._file_path, jdf)`

src/main/python/smv/smviostrategy.py
ninjapapa: will do.

ninjapapa:
Changes made as required.

ninjapapa:
@AliTajeldin ready to go?

PR #1455, opened 2018-10-29
I1221 hot load flg

### LINKED ISSUE #1221: Option to turn off hotloading of Python modules
Reloading modules leads to annoying gotchas when writing custom runners.

ninjapapa:
Fixed #1221


src/main/python/smv/smvdriver.py
```diff
@@ -21,7 +21,9 @@ def create_smv_app(self, smv_args, driver_args):
                 smv_args (list(str)): CLI args for SMV - should be passed to `SmvApp`)
                 driver_args (list(str)): CLI args for the driver
         """
-        return SmvApp.createInstance(smv_args)
+        # When SmvDriver is in use, user will call smv-run and interact
+        # through command-line, so no need to do py module hotload
+        return SmvApp.createInstance(smv_args, py_module_hotload=False)
```
AliTajeldin: @laneb Does Eddie user `SmvDriver` or creates instance directly?

src/test/python/testSmvFramework.py
```diff
@@ -242,3 +245,50 @@ def test_publish_as_file(self):
...
+        return ['--smv-props', 'smv.stages=hotload', '-m', "None"]
+
+    def test_with_hotload(self):
+        """Module M1 should be loaded twice"""
+        global module_load_count
+        module_load_count = 0
+        fqn = "hotload.modules.M1"
+        self.load(fqn)
+        self.load(fqn)
+        self.assertEqual(module_load_count, 2)
```
laneb: Should add a note explaining how `module_load_count` is getting incremented.

src/test/python/testSmvFramework.py
```diff
@@ -242,3 +245,50 @@ def test_publish_as_file(self):
...
+
+class SmvAppNoPyHotLoadTest(SmvBaseTest):
+    @classmethod
+    def smvAppInitArgs(cls):
+        return ['--smv-props', 'smv.stages=hotload', '-m', "None"]
+
+    @classmethod
+    def setUpClass(cls):
+        # Import needs to happen during EVERY setup to ensure that we are
+        # using the most recently reloaded SmvApp
```
laneb: Not clear what this means. Why would `SmvApp` ever be reloaded? SMV doesn't reload its own source.

src/test/python/testSmvFramework.py
```diff
@@ -242,3 +245,50 @@ def test_publish_as_file(self):
...
+        cls.mkTmpTestDir()
+
+    def test_without_hotload(self):
+        """Module M1 should only be loaded once"""
+        global module_load_count
+        module_load_count = 0
+        fqn = "hotload.modules.M1"
+        self.load(fqn)
+        self.load(fqn)
+        self.assertEqual(module_load_count, 1)
```
laneb: This assumes that `hotload.modules` wasn't cached before this test was run. If it was cached then the line that increments `module_load_count` wouldn't run even once. Should clear `sys.modules` in the test setup so we're not relying on test order.

PR #1458, opened 2018-10-30
I1456 parser err handling

### LINKED ISSUE #1456: Refine the parser error handling logic of all SmvInput with parsers
Currently we have 2 flags:

* failAtParsingError - add a `FailParserCountPolicy(1)` policy to users DQM policy
* forceParserCheck - label the DQM as "need an action"

The "need an action" logic is no more used, since if we call `runModule` without a quick run flag, an action anyhow will be triggered. In case we cam `runModule` on the input module itself, an action will be forced with or without the `forceParserCheck` flag.

There is another place we can change the the behavior of handling parser error. When we call `doRun` method, it need a `validator` parameter, and if the validator is None, a `TerminateParserLogger` will be used. In that case, when there is any parser error, the executor will throw an exception, and the `TerminateParserLogger` will not capture and log into an accumulator as normal, instead, it will just throw that exception, and stop.

However, the `doRun` method is not called with None in any cases for SmvInput with parsers. 

This issue will:
* Remove `forceParserCheck` flag
* Make `failAtParsingError` use `TerminateParserLogger`, so that it will fail early, for user who want to tolerant some, they should set `failAtParsingError` to false, and add some parsing policy
* Document the best practice

ninjapapa:
Fixed #1456


src/main/python/smv/smvinput.py
```diff
@@ -358,7 +351,7 @@ def readAsDF(self, readerLogger):
                 filePath,
                 self.smvSchema(),
                 self.csvAttr(),
-                readerLogger
+                self.readerLogger()
```
AliTajeldin: isnt this creating a new logger per file?

ninjapapa: Yes. But `ParserLogger` is a very small class:
https://github.com/TresAmigosSD/SMV/blob/master/src/main/scala/org/tresamigos/smv/dqm/ParserLogger.scala

Is there a concern of GC on them?

ninjapapa: Could move them out of the loop anyhow.

ninjapapa: Oh I see, it created multiple dqmStates in that case. 
Fixing.

PR #1474, opened 2018-11-03
I1469 meta closer to run

### LINKED ISSUE #1469: metadata may not be created while data persisted successfully
Call `get_data` and `persist_meta` are two different visits (actually more then 2), it will happen much more often that when a long running job get killed, there will have modules which have valid persisted data, but no persisted metadata.

One way to fix is by bring the persist_meta call in the same visit to the get_data for all the non-ephemeral modules. For ephemeral modules, it is not a problem, since anyhow there are no data persisted.

ninjapapa:
Fixed #1469 

Moved meta calculation, validation, persisting, hist update & persisting all into run_ancestor_and_me_postAction method. 

The run method of runner reduced to 2 visits, one for run, one for handle leftover ephemeral post actions.

### DIFF
--- a/src/main/python/smv/smvgenericmodule.py
+++ b/src/main/python/smv/smvgenericmodule.py
@@ -17,8 +17,8 @@ import binascii
 import json
 from datetime import datetime
 
-from smv.utils import lazy_property
-from smv.error import SmvRuntimeError
+from smv.utils import lazy_property, is_string
+from smv.error import SmvRuntimeError, SmvMetadataValidationError
 from smv.modulesvisitor import ModulesVisitor
 from smv.smvmetadata import SmvMetaData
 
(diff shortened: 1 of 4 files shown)


src/main/python/smv/smvgenericmodule.py
```diff
@@ -455,21 +465,51 @@ def get_metadata(self):
...
-            self.run_ancestor_and_me_postAction(run_set)
+            self.run_ancestor_and_me_postAction(run_set, collector)
 
-    def run_ancestor_and_me_postAction(self, run_set):
+    def run_ancestor_and_me_postAction(self, run_set, collector):
         """When action happens on current module, run the the delayed 
             post action of the ancestor ephemeral modules
         """
-        def run_delayed_postAction(mod, _run_set):
+        def not_persisted_or_no_edd_when_forced(io_strategy):
```
AliTajeldin: great function name.  Makes the code below much easier to read.

PR #1480, opened 2018-11-05
I1473 lock conf

### LINKED ISSUE #1473: Lock files should be in a separate dir and turn on/off through config
Create $dataDir/lock dir, and put all lock files there. Also introduce `smv.lockDir` and `smv.lock` options.

ninjapapa:
Fixed #1473

### DIFF
--- a/src/main/python/smv/smviostrategy.py
+++ b/src/main/python/smv/smviostrategy.py
@@ -70,15 +70,6 @@ class SmvFileOnHdfsIoStrategy(SmvIoStrategy):
         else:
             self._file_path = file_path
 
-        self._lock_path = self._file_path + ".lock"
-
-    def _smvLock(self):
-        # get a lock for 1 hour
-        return self.smvApp._jvm.org.tresamigos.smv.SmvLock(
-            self._lock_path,
-            3600 * 1000 
-        )
-    
     @abc.abstractmethod
     def _read(self):
         """The raw io read action"""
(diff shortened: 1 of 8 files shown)


src/main/python/smv/smvlock.py
```diff
@@ -0,0 +1,38 @@
...
+            self._lock_path,
+            3600 * 1000 
+        )
+        self.slock.lock()
+        return None
+
+    def __exit__(self, type, value, traceback):
+        self.slock.unlock()
+
+class NonOpLock(object):
```
AliTajeldin: neato

src/main/python/smv/smviostrategy.py
```diff
@@ -92,19 +83,8 @@ def _write(self, raw_data):
...
-            else:
-                self.smvApp.log.info("No cached result found. Caching result at {}".format(self._file_path))
-                # Delete outputs in case data was partially written previously
-                # since `isPersisted` test on schema file, this case only happens when schema was written half way
-                self.remove()
-                self._write(dataframe)
-        finally:
-            slock.unlock()
+        # May add lock or other logic here in future
+        self._write(dataframe)
```
AliTajeldin: where do we acquire the lock for the data write?

ninjapapa: Current file based lock is acquired from the caller. Maybe in the future the write itself has some other way to lock, maybe not. Should we just remove `_write` and `_read` and use the one without underscore?

AliTajeldin: Yah, I like that better as `_xxx` implies that `xxx` is somewhat private (by convention)

AliTajeldin: Let's discuss location of lock/unlock tomorrow.  Would be in another ticket if we need to change anyway.  Merging this.

PR #1492, opened 2018-11-09
I1491 smvjdbcio

### LINKED ISSUE #1491: Implement SmvJdbcInputTable and SmvJdbcOutputTable
This issue is to setup the `SmvIoModule` framework with a single connection and input/output type to proof of concept the design in #1484

ninjapapa:
Fixed #1491 

## Summary

* Separated `SmvGenericModule` to `SmvGenericModule <- SmvProcessModule`
* Implemented `SmvGenericModule <- SmvIoModule` and `SmvInput`/`SmvOutput`
* For `SmvOutput`, current implementation can meet the need to existing mixin use case, and new IO framework
* Implemented `SmvConnectionInfo` with a single example `SmvJdbcConnectionInfo`
* Implemented a single IO type: `SmvJdbcInputTable`/`SmvJdbcOutputTable`


src/main/python/smv/smviomodule.py
```diff
@@ -0,0 +1,201 @@
...
+    @abc.abstractmethod
+    def connectionName(self):
+        """Name of the connection to read/write"""
+    
+    def get_connection(self):
+        """Get connection instance from name"""
+        name = self.connectionName()
+        props = self.smvApp.py_smvconf.merged_props()
+        type_key = "smv.con.{}.type".format(name)
+        con_type = props.get(type_key).lower()
```
laneb: If the type doesn't exist in the conf, this will result in a `NoneType has no attribute...`.

src/main/python/smv/smviomodule.py
```diff
@@ -0,0 +1,201 @@
...
+    def instanceValHash(self):
+        """TODO: need to implement this to depends on connection and 
+            also table itself"""
+        return 0
+
+class SmvOutput(SmvIoModule):
+    """Reuse the output marker mixin, should work for old mixin and new base class.
+        Will make it symmetric to SmvInput when mixin usage is deprecated.
+        
+        Mixin which marks an SmvModule as one of the output of its stage
```
laneb: This should be the docstring header. The note about reusing the marker is useful historical context but doesn't help explain what `SmvOutput` is/does.

src/main/python/smv/smviomodule.py
```diff
@@ -0,0 +1,201 @@
...
+                (string)
+        """
+        # Implemented as None to be backword compatable as the mixin marker
+        return None
+
+    def connectionName(self):
+        # Implemented as None to be backword compatable as the mixin marker
+        return None
+    
+    def assert_single_input(self):
```
laneb: `SmvOutput` only needs to be backwards compatible enough for Atlas use right? This will definitely break T's project.

ninjapapa: assert_single_input is not called, if `SmvOutput` is used for mix in.

src/main/python/smv/smvconnectioninfo.py
```diff
@@ -0,0 +1,74 @@
...
+            Each connection must have a "smv.con.{connection_name}.type" prop.
+            For different connections, other props may exists for the attributes
+        """
+        prop_prefix = "smv.con.{}.".format(self.name)
+        for a in self.attributes():
+            prop_key = prop_prefix + a
+            setattr(self, a, props.get(prop_key, None))
+
+
+class SmvJdbcConnectionInfo(SmvConnectionInfo):
```
AliTajeldin: This should be in a different file.

src/main/python/smv/smvconnectioninfo.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def attributes(self):
+        return ["url", "driver", "user", "password"]
+
+
+_all_connections = [
+    SmvJdbcConnectionInfo,
+]
+
+def getConnection(con_type):
```
AliTajeldin: should be `getConnectionInfo`

src/main/python/smv/smvconnectioninfo.py
```diff
@@ -0,0 +1,74 @@
...
+class SmvJdbcConnectionInfo(SmvConnectionInfo):
+    @classmethod
+    def connection_type(cls):
+        return "jdbc"
+
+    def attributes(self):
+        return ["url", "driver", "user", "password"]
+
+
+_all_connections = [
```
AliTajeldin: Hmm, need to find a way to allow other external connection info classes to register.
Wish python had something similar to java service provider interface.
One option is to do this in config.  Allow user to specify the connection info class associated with connection type.  For example:
```
smv.con.type.jdbc.class = SmvJdbcConnectionIfno
```

Then when a user defines a connection of type `X`, we lookup `smv.con.type.X.class` for the actual class and we wouldn't need to keep a list of all known connection info types.

src/main/python/smv/smviomodule.py
laneb: Got it, because `SmvOutput` is not overriding `doRun`.

src/main/python/smv/smvconnectioninfo.py
```diff
@@ -40,6 +40,11 @@ def __init__(self, name, props):
         self._from_props(props)
 
     def _from_props(self, props):
+        """Create an SmvConnectionInfo instance from smv props
```
laneb: Ought to be clear that this method is loading the connection attributes for an instance, not creating an instance.

src/main/python/smv/smvconnectioninfo.py
ninjapapa: In that case, should we even need to specify "type"? In other words:

```
smv.con.my_jdbc.class = SmvJdbcConnectionInfo
smv.con.my_jdbc.driver = ...
...
```

Then `getConnectionInfo` method with get a string of `SmvJdbcConnectionInfo`, and it should be able to search `sys.path` to load that class.

ninjapapa: Another way to do this is to make the list in a separate python file, and allow user to append it, just like `sys.path`.

laneb: In the `smv.con.my_jdbc.class = SmvJdbcConnectionInfo`, the user is providing `fqn` not just basename, correct?

AliTajeldin:
Will be merged once we get 👍 from @laneb

ninjapapa:
@laneb ready to merge?

src/main/python/smv/smviomodule.py
```diff
@@ -47,10 +48,12 @@ def get_connection(self):
...
         name = self.connectionName()
         props = self.smvApp.py_smvconf.merged_props()
-        type_key = "smv.con.{}.type".format(name)
-        if (type_key in props):
-            con_type = props.get(type_key).lower()
-            return getConnectionInfo(con_type)(name, props)
+        class_key = "smv.con.{}.class".format(name)
+        if (class_key in props):
+            con_class = props.get(class_key)
+            module_name, class_name = con_class.rsplit(".", 1)
```
laneb: That's a useful idiom!

src/main/python/smv/smviomodule.py
```diff
@@ -47,10 +48,12 @@ def get_connection(self):
...
+            ConnClass = getattr(importlib.import_module(module_name), class_name)
+            return ConnClass(name, props)
         else:
             raise SmvRuntimeError("Connection name {} is not configured with a type".format(name))
```
laneb: type -> class to match the config schema

PR #1493, opened 2018-11-10
I958 ignore smv shell all args

### LINKED ISSUE #958: Add .smv_shell_all_args to project templates' gitignore
SMV copies the `smv_pyshell_init.py` into my project directory. Because it does not pass any linting or type-checking, it breaks our current CI tooling. 

While I think the long-term solution is to have our bash scripts and python scripts pass common linting and type-checking tools, this is expensive and we could never anticipate what tooling any given user might have. 

I think the best fix is to make this file a dotfile, and therefore consider it more of a private implementation detail than a publicly exposed file. Most clients won't even be aware that it exists then, and most tooling should ignore it by default. 

If we do plan on making this file more of a first-class citizen in our interpreter environment setup routine, then we might want to consider at least documenting its presence and purpose somewhere, and force the user to check it into source control and handle it in her tooling.

ninjapapa:
Fixed #958


ninjapapa:
Merge as it is very simple changes.

PR #1500, opened 2018-11-14
I1499 io sub pack

### LINKED ISSUE #1499: Put smviomodule to a sub-package
To reduce chance of conflicts with exiting input and output modules, will put the new IO modules in a sub-package of smv.

ninjapapa:
Fixed #1499 

* Moved new io modules to smv.iomod sub-package
* Get back the SmvOutput mixin (it's better to keep it separated from the base class smv.iomod.SmvOutput)


src/main/python/smv/iomod/base.py
```diff
@@ -151,85 +118,32 @@ def assert_single_input(self):
...
-class SmvJdbcOutputTable(SmvOutput):
-    """
-        User need to implement 
-
-            - requiresDS
-            - connectionName
-            - tableName
-            - writeMode: optional, default "errorifexists"
-    """
+class AsTable(object):
```
jacobdr: Might want to consider a better name than `AsTable`... Maybe `TableLike` or `ImplementsTable`?

ninjapapa: Agree need a better name, but either `TableLike` nor `ImplementsTable` works. Since this class is "private" (in the sense `from smv.iomod import *` will not import it), may be we should prefix with "_"?

src/main/python/smv/iomod/inputs.py
```diff
@@ -0,0 +1,44 @@
...
+class SmvJdbcInputTable(SmvInput, AsTable):
+    """
+        User need to implement 
+
+            - connectionName
+            - tableName
+    """
+
+    def doRun(self, known):
+        conn = self.get_connection()
```
AliTajeldin: The actual reading/writing of jdbc data should be in the JDBC IO strategy.

Issue #1503, opened 2018-11-14
Implement core SmvInput using the new iomod interface and deprecate some rarely used ones


ninjapapa:

ninjapapa:
Steps:
* [ ] move XML to iomod (no need to leave one in smvinput.py)
* [ ] remove jdbc from smvinput and move hdfs to iomod, but keep a wrapper in smvinput.py for backward compatability
* [ ] CSV and MultiCsv
* [ ] CSV String data

PR #1508, opened 2018-11-16
I1497 csv io

### LINKED ISSUE #1497: Implement csv i/o files
For the new `SmvCsvInputFile`, the key question is how to handle Schema files.

### Option 1
Assume schema files and csv files are always in the same connection. If not, at project level, provide a `smv.schemaDir` location to store the schema files. So `SmvCsvInputFile` will try the csv connection first, then the global `schemaDir` (or the other order).

### Option 2
`SmvCsvInputFile` has 2 connections, one csv, one schema. Default still assume schema is in the same connection as csv, but allow user to specify a schema connection, and schema relative path.

ninjapapa:
Fixed #1497 

Using the new iomod package to implement csv input and output. Some code was copied from `smvinput.py`. The original version will be removed in issue #1503.

### DIFF
--- a/src/main/python/smv/iomod/inputs.py
+++ b/src/main/python/smv/iomod/inputs.py
@@ -100,7 +100,7 @@ class InputFileWithSchema(SmvInput, AsFile):
             return self.schemaFileName()
         else:
             return self.fileName().rsplit(".", 1)[0] + ".schema"
-        
+
     def get_schema_string(self):
         """Return the schema string specified by user either through
             userSchema method, or through a schema file. The priority is the following:
@@ -168,8 +168,64 @@ class SmvCsvInputFile(InputFileWithSchema, WithCsvParser):
         )
         return DataFrame(jdf, self.smvApp.sqlContext)
 
+
+class SmvMultiCsvInputFiles(InputFileWithSchema, WithCsvParser):
+    # Override schema_file_name logic
+
+    @abc.abstractmethod
+    def dirName(self):
+        """Path to the directory containing the csv files
+            relative to the path defined in the connection
+
+            Returns:
+                (str)
+        """
+
+    def _get_schema_file_name(self):
+        """The schema_file_name is determined by the following logic
+
+                - schemaFileName
+                - dirName with post-fix schema
+        """
+        if (self.schemaFileName() is not None):
+            return self.schemaFileName()
+        else:
+            return self.dirName() + ".schema"
+
+    def fileName(self):
+        return None
+
+    def doRun(self, known):
+        dir_path = "{}/{}".format(self.get_connection().path, self.dirName())
+        smvSchemaObj = self.smvApp.j_smvPyClient.getSmvSchema()
+        smv_schema = smvSchemaObj.fromString(self.get_schema_string())
+        csv_attr = smv_schema.extractCsvAttributes()
+
+        flist = self.smvApp._jvm.SmvHDFS.dirList(dir_path).array()
+        # ignore all hidden files in the data dir
+        filesInDir = ["{}/{}".format(dir_path, n) for n in flist if not n.startswith(".")]
+
+        if (not filesInDir):
+            raise SmvRuntimeError("There are no data files in {}".format(dir_path))
+
+        combinedJdf = None
+        reader_logger = self.readerLogger()
+        for filePath in filesInDir:
+            jdf = self.smvApp.j_smvPyClient.readCsvFromFile(
+                filePath,
+                smv_schema,
+                csv_attr,
+                reader_logger
+            )
+            combinedJdf = jdf if (combinedJdf is None) else combinedJdf.unionAll(jdf)
+
+        return DataFrame(combinedJdf, self.smvApp.sqlContext)
+
+
+
 __all__ = [
     'SmvJdbcInputTable',
     'SmvHiveInputTable',
     'SmvCsvInputFile',
+    'SmvMultiCsvInputFiles',
 ]
--- a/src/main/python/smv/iomod/inputs.py
+++ b/src/main/python/smv/iomod/inputs.py
@@ -81,6 +81,9 @@ class InputFileWithSchema(SmvInput, AsFile):
             - User specified in current module through schemaConnectionName method
             - Configured in the global props files with prop key "smv.schemaConn"
             - Connection for data (user specified through connectionName method)
+
+            Since in some cases user may not have write access to the data folder,
+            need to provide more flexibility on where the schema files can come from.
         """
         name = self.schemaConnectionName()
         props = self.smvApp.py_smvconf.merged_props()
--- a/src/main/python/smv/iomod/inputs.py
+++ b/src/main/python/smv/iomod/inputs.py
@@ -11,9 +11,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from smv.iomod.base import SmvInput, AsTable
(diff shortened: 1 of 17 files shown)


src/main/python/smv/iomod/inputs.py
```diff
@@ -41,7 +47,204 @@ def doRun(self, known):
...
+            - fileName: required
+            - schemaConnectionName: optional
+            - schemaFileName: optional
+            - userSchema: optional
+            - failAtParsingError: optional, default True
+            - dqm: optional, default SmvDQM()
+    """
+
+    def doRun(self, known):
+        file_path = "{}/{}".format(self.get_connection().path, self.fileName())
```
laneb: We should tolerate if the connection path terminates with a '/'. `os.path.join` lets us not think about those details.

src/main/python/smv/iomod/inputs.py
```diff
@@ -41,7 +47,204 @@ def doRun(self, known):
...
+    @abc.abstractmethod
+    def dirName(self):
+        """Path to the directory containing the csv files
+            relative to the path defined in the connection
+
+            Returns:
+                (str)
+        """
+
+    # Override schema_file_name logic
```
laneb: Comment doesn't explain why.

src/main/python/smv/iomod/inputs.py
```diff
@@ -41,7 +47,204 @@ def doRun(self, known):
...
+        if (self.schemaFileName() is not None):
+            return self.schemaFileName()
+        else:
+            return self.dirName() + ".schema"
+
+    def fileName(self):
+        return None
+
+    def doRun(self, known):
+        dir_path = "{}/{}".format(self.get_connection().path, self.dirName())
```
laneb: Same point about using `os.path.join`

src/main/python/smv/iomod/outputs.py
```diff
@@ -72,7 +75,28 @@ def doRun(self, known):
...
+    """
+        User need to implement
+
+            - requiresDS
+            - connectionName
+            - fileName
+    """
+    def doRun(self, known):
+        data = self.get_spark_df(known)
+        file_path = "{}/{}".format(self.get_connection().path, self.fileName())
```
laneb: `os.path.join`

src/main/python/smv/iomod/outputs.py
```diff
@@ -72,7 +75,28 @@ def doRun(self, known):
...
+        schema_path = re.sub("\.csv$", ".schema", file_path)
+
+        smvSchemaObj = self.smvApp.j_smvPyClient.getSmvSchema()
+        schema = smvSchemaObj.fromDataFrame(data._jdf, "_SmvStrNull_", self.smvApp.scalaOption(CsvAttributes()))
```
laneb: Can we avoid hard coding `_SmvStrNull_` here? It should be declared as a constant somewhere.

src/main/python/smv/iomod/inputs.py
```diff
@@ -41,7 +47,204 @@ def doRun(self, known):
...
+    def _get_schema_file_name(self):
+        """The schema_file_name is determined by the following logic
+
+                - schemaFileName
+                - fileName replace the post-fix to schema
+        """
+        if (self.schemaFileName() is not None):
+            return self.schemaFileName()
+        else:
+            return self.fileName().rsplit(".", 1)[0] + ".schema"
```
laneb: If the file is the located in something like "/data/foo.project/input/mycsv" this won't produce the desired file name (`/data/foo.project/input/mycsv.schema`). IMO we should enforce file names ending in `.csv` unless there's a strong reason not to.

src/main/python/smv/smviostrategy.py
```diff
@@ -297,3 +302,42 @@ def write(self, raw_data):
 
     # TODO: we should allow persisting intermidiate results in Hive also
     # For that case, however need to specify a convention to store semaphore
+
+
+class SmvSchemaOnHdfsIoStrategy(SmvIoStrategy):
+    """Simple read/write a text file on Hdfs"""
```
laneb: Docstring is inaccurate.

src/main/python/smv/smviostrategy.py
```diff
@@ -297,3 +302,42 @@ def write(self, raw_data):
...
+        self.smvApp = smvApp
+        self._file_path = path
+
+    def read(self):
+        schema_file_str = self.smvApp._jvm.SmvHDFS.readFromFile(self._file_path)
+        smvSchemaObj = self.smvApp.j_smvPyClient.getSmvSchema()
+        smv_schema = smvSchemaObj.fromString(";".join(schema_file_str.split("\n")))
+        return smv_schema
+
+    def write(self, rawdata):
```
laneb: Should note that the expected type of rawdata in an `SmvSchema` instance.

src/main/python/smv/iomod/inputs.py
```diff
@@ -41,7 +47,204 @@ def doRun(self, known):
...
+    def dqmValidator(self):
+        return self.smvApp._jvm.DQMValidator(self.dqm())
+
+    def readerLogger(self):
+        if (self.failAtParsingError()):
+            return self.smvApp._jvm.SmvPythonHelper.getTerminateParserLogger()
+        else:
+            return self.dqmValidator.createParserValidator()
+
+class WithSmvSchema(InputFileWithSchema):
```
laneb: Do these need to be separate classes? Will `InputFileWithSchema` be used elsewhere without `WithSmvSchema`?

ninjapapa:
Schema file will be potential in a place different than csv files. Please see the docstr of `_get_schema_connection` method. Basically we will allow 4 different ways for user to specify schema of a CSV file:

* In case that the CSV will only be used in a single SMV project, user can code the schema into code through `userSchema` method
* In case that a CSV might be shared across multiple projects, need to put schema to a file. There are 3 places can store the schema file:
    - If CSV data folder is writable, user can put discovered schema file beside the data file
    - If CSV data folder is not writable, a project level global schema folder can be specified through app config
    - If CSV data's schema is in a specific place which might be different from the global schema folder, user can specify through `schemaConnectionName` for that special location

This flexibility is needed since some user may want schema to be closer to data, some others may not or may not be allowed.

src/main/python/smv/iomod/inputs.py
ninjapapa: Will move content of `WithSmvSchema` to `SmvCsvInputFile`. `InputFileWithSchema` can also be used by XML input data, which will be added soon.

ninjapapa: Can't move to `SmvCsvInputFile`, since `SmvMultiCsvInputFiles` is also using it.

src/main/python/smv/iomod/outputs.py
ninjapapa: How about introduce a conf parameter `smv.strnull` and default to `_SmvStrNull_`? We can create a new issue for this change.

laneb: I like it.

src/main/python/smv/iomod/inputs.py
ninjapapa: That is right. XML's schema will be managed as pure spark schema, and persisted as JSON.

ninjapapa: The difference is "schema" vs. "SmvSchema".

PR #1510, opened 2018-11-19
I1509: use python to initialize git repo

### LINKED ISSUE #1509: Use python package to initialize git repository
Python package `dulwich` provides an interface to git repos that doesn’t call out to git directly but instead uses pure Python.

fei-cheng:
Fixes #1509 

The `git log` of the repo which initialized by using python looks like:
```
eddie@b55c706dfa6d:~/testInit$ git log
commit 2809c753789d9f56161c1bda82fbb79a1f4d2e2b
Author: eddie <eddie@b55c706dfa6d>
Date:   Mon Nov 19 08:35:07 2018 +0000

    init git repository
```

### DIFF
--- a/docker/smv/requirements.txt
+++ b/docker/smv/requirements.txt
@@ -1,3 +1,4 @@
+dulwich
 Flask
 # iPython 6+ does not support Python < 3.3, so need to pin this for Python2 support
 # https://ipython.readthedocs.io/en/stable/whatsnew/version5.html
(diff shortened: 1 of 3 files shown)


laneb:
This doesn't belong in this repo. We should probably just get rid of `smv-init's` `--git` option.

AliTajeldin:
agree with @laneb .  Need to close this PR.  Just make sure that `smv-init` always has an option to run without doing a `git init`.

PR #1511, opened 2018-11-21
I1503 step1 xml

ninjapapa:
Step1 of #1503 

Moved `SmvXmlFile` to `iomod. SmvXmlInputFile`


src/main/python/smv/iomod/inputs.py
```diff
@@ -106,6 +110,49 @@ def _get_schema_file_name(self):
...
+            s_path = self._get_schema_file_name()
+            try:
+                s = SmvTextOnHdfsIoStrategy(self.smvApp, s_path).read()
+                return str_to_schema(s)
+            except:
+                return None
+
+    def doRun(self, known):
+        """readin xml data"""
+        file_path = "{}/{}".format(self.get_connection().path, self.fileName())
```
laneb: `os.path.join`

PR #1515, opened 2018-11-26
I1503 step2 jdbc hive

ninjapapa:
Step 2 of #1503 

Since `SmvJdbcTable` in `smvinput` was rarely used in projects, simply removed it (replaced by the `iomod.SmvJdbcInputTable`).

`SmvHiveTable` was used in multiple projects, so still keep for now. However "query-on-read" is no more supported.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,8 +78,12 @@ setuptools.setup(
     python_requires=REQUIRES_PYTHON,
     url=URL,
     tests_require=['sphinx', 'tox'],
-    packages=[
-        'smv',
+    # Need to call find_packages so that newly introduced
+    # sub-packages will be included
+    packages=setuptools.find_packages(
+        "src/main/python",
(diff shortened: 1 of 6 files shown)


ninjapapa:
Integration test fail. Close to fix.
```
ERROR: InvocationError for command '/bin/bash src/test/scripts/run-integration-test.sh --pip-install' (exited with code 1)
```

setup.py
```diff
@@ -78,8 +78,12 @@ def read_file(path_from_root):
     python_requires=REQUIRES_PYTHON,
     url=URL,
     tests_require=['sphinx', 'tox'],
-    packages=[
-        'smv',
+    # Need to call find_packages so that newly introduced
+    # sub-packages will be included
+    packages=setuptools.find_packages(
```
AliTajeldin: neato

PR #1516, opened 2018-11-26
I1512 jvmless logging

### LINKED ISSUE #1512: Switch to a Python library for logging
When I added logging to SMV earlier this year, I used log4j since it was readily available and could be accessed both Scala and from Python (using Py4J). After discussing performance issues and optimizations re: graph generation with @AliTajeldin, I wanted to see where our bottleneck(s) are. It turns out logging through the JVM is a big bottleneck - dropping in a Python standard logger in place of Py4J for logging from the Python side of the application improved introspection performance by an order of magnitude. When running a trivial module in a 1000 module project, the running time is about 15s and is dominated introspection. After replacing the logging, the running time drops to about 3s. So dropping log4j will significantly speed up introspection-dominated tasks.

laneb:
Resolves #1512.

Replaces log4j logger with Python standard logger for all logging in SMV:
* Create new logger using Python standard `logging` library and make it available for import at `smv.logger`
* Migrate existing log statements to using `smv.logger` instead of the `SmvApp` logger
* Move final remaining log statement from Scala to Python using the new logger

Removing the cost of py4j from logging improves the latency of `get_graph_json` for a 1000 module project by over 50%. For now logging levels and other configuration will need to be adjusted through code (e.g. `smv.logger.setLevel("DEBUG")`). `logging` supports configuration via yaml, so I created #1517 to load a custom config provided by users.

### DIFF
--- a/docs/user/dqm.md
+++ b/docs/user/dqm.md
@@ -41,8 +41,7 @@ class Myfile(smv.SmvCsvFile):
 
 With above setting, the `SmvCsvFile` will simply persist the validation result and keep moving.
 
-Either terminating the process or not, as long as the log is nontrivial, it will be logged in
-log4j with level "warning".
+Either terminating the process or not, as long as the log is nontrivial, it will be logged with level "warning".
 The result will also by persisted in the `SmvModule` persisted metadata with postfix `.meta`.
 
 Sometimes we need more flexibility on specifying the terminate criteria. For example, I can tolerate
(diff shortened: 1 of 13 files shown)


laneb:
Missed that `smv.dsm.TX` was logging through log4j using a different strategy. Will migrate that to use `smv.logger` as well.

laneb:
* Short term, user sets log level and other logging configuration programmatically (e.g. `smv.logger.setLevel("DEBUG")`). I created #1517 to let SMV load logging conf from file using `logging.config.fromFile`. 

* Tying logging to `SmvApp`	requires the injecting the `SmvApp` (whether by constructor or by `SmvApp.getInstance`) into places where it is otherwise not needed. It will be a lot simpler to add new logging to an arbitrary function if it doesn't need to be tied to `SmvApp`. Could leave the existing log statements going through `SmvApp`, but didn't want to maintain 2 patterns for logging. By "less changes" do you mean to reduce the size of the PR?

AliTajeldin:
Got it, makes sense.  merging.

Issue #1498, opened 2018-11-13
provide something similar to Java Service Provider API in SMV


AliTajeldin:
# Problem
We have run into multiple issues now with solutions that require a way to "register" a set of classes that provide a type of service and then query for the list of all known such classes.  For example, for connection classes, we would like to know all the known available connection classes so we can map the user specified type to the actual class type.

# Solution
Need to create a **lightweight** solution (so no distributed consul, zookeeper, or other large scale external solutions) that is embedded within SMV.  The proposal is to use python decorators (https://realpython.com/primer-on-python-decorators/#decorating-classes) to allow users to easily declare providers and their associated metadata.

## Decorator
SMV will provide the `SmvProvider` class decorator that users can utilize to declare their provider classes/interfaces.  For example:
```python
@SmvProvider("conn", "jdbc")
class SmvJdbcConnectionInfo(SmvConnectionInfo):
   def attributes(): return ["url", ...]

  @abstract
  def url():
```

The `SmvProvider` takes two and only two arguments.  The type of the service being provided and the name of this provider.  In the example above, the type is `conn` and the name is `jdbc`.  Users would be able to ask SMV for list of all known `conn` providers and so on.

**Note:** for the registration to work, the above code must be "executed".  The enclosing python module must be imported someplace.  The easiest way to do this is to enumerate all the provider files into a single "index" file and import the index file so force the import of all the provider files.  SMV will do that automatically for all its known providers (e.g. `SmvHdfsDirConnection`)

## Metadata
Rather than burden the provider registry with metadata, we should create a base provider class for each known provider type and declare abstract methods in the base class to force the concrete classes to provide the required metadata.  The provider interface should be kept simple and minimal.

For example, assume we have a provider type X that needs a foo attribute for each concrete provider. Instead of adding extra `foo` parameter to the `SmvProvider` decorator, we should create a `BaseX` class that declares `foo()` method as abstract.  So:
**WRONG**
```python
@SmvProvider("X", "myX", foo=15)
class MyX(Object):
  ...
```

**CORRECT**
```python
class BaseX(Object)
  @abstract
   def foo():

@SmvProvider("X", "myX")
class MyX(BaseX):
  def foo(): return 15
```

# TBD
* dyanmic loading.  Should these classes ever be removed from `sys.modules` cache and always reloaded or should we assume they don't change.
*

laneb:
Couple questions:
[...]

PR #1518, opened 2018-11-27
#1498: initial smv provider impl

AliTajeldin:
phase 1 of #1498 
This is just for the discovery part.  Had to refactor some of the module discovery code to reuse the code for the discovery of providers.
The next phase will add support for getting the provider full name, etc.

### DIFF
--- a/src/main/python/smv/datasetrepo.py
+++ b/src/main/python/smv/datasetrepo.py
@@ -163,6 +163,10 @@ class DataSetRepo(object):
             obj = getattr(pymod, obj_name)
             self.smvApp.log.debug("Inspecting {} ({})".format(obj_name, type(obj)))
 
+            # skip non-class objects
+            if not inspect.isclass(obj):
+                continue
+
             if not is_matching(obj):
                 self.smvApp.log.debug("Ignoring {} because it is not a match.".format(obj_name))
                 continue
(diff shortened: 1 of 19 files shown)


src/main/python/smv/datasetrepo.py
```diff
@@ -146,11 +145,74 @@ def onerror(name):
...
+        # iterate over the attributes of the module, looking for SmvGenericModules
+        for obj_name in dir(pymod):
+            obj = getattr(pymod, obj_name)
+            self.smvApp.log.debug("Inspecting {} ({})".format(obj_name, type(obj)))
+
+            if not is_matching(obj):
+                self.smvApp.log.debug("Ignoring {} because it is not a match.".format(obj_name))
+                continue
+
+            obj_fqn = obj.__module__ + "." + obj.__name__
```
laneb: This will fail if `obj` is e.g. a `str` or an `int` constant, which have neither `__module__` nor `__name__`. `_matchingClassesInPyModule` is currently relying on `is_matching to filter out non-classes. `_matchingClassesInPyModule` should check that explicitly. Can use `inspect.isclass`.

AliTajeldin:
@laneb 
* good catch, added the check for `isclass`.
* `SmvGenericModule` will not be a provider but some derived class probably will (e.g. `SmvSparkLRModel`)
* merged with master to remove conflict.

Ready for re-review.

PR #1523, opened 2018-11-30
 #1521 created get_module_graph_json

### LINKED ISSUE #1521: Drop needsToRun from get_graph_json method in smvapp
Current `get_graph_json` method has module state (needToRun). Need to create a method which just return the module dependency graph without the module state info. Will name that method as `get_module_graph_json`. 

As client code converted to use:

* `get_module_graph_json`, and
* `get_module_state_json`

will deprecate `get_graph_json` method.

ninjapapa:
Fixed #1521


ninjapapa:
Hold this for a rethink of the api design

PR #1527, opened 2018-12-01
I1503 step3 csv

### LINKED ISSUE #1503: Implement core SmvInput using the new iomod interface and deprecate some rarely used ones

ninjapapa:
Fixed #1503 

* Separated `SparkDfGenMod` so that both `SmvSparkDfModule` and all input modules which create DF can mix it in
* Made all CSV input modules in `smvinput` derived from `iomod` modules
* Fixed tests


ninjapapa:
itest failed. Close to fix.

ninjapapa:
Itest was failed because `instanceValHash` was alway 0. Fixed.

PR #1528, opened 2018-12-01
 #1521 made get_module_state_json takes a list

ninjapapa:
Step 1 of #1521


src/main/python/smv/smvappinfo.py
```diff
@@ -92,10 +92,11 @@ def edge_pair(f, t):
             "edges": [edge_pair(p[0], p[1]) for p in edges]
         })
 
-    def create_module_state_json(self):
+    def create_module_state_json(self, fqns):
         """Create all modules needToRun state Json string
         """
-        nodes = self.dsm.allDataSets()
+        urns = ["mod:" + fqn for fqn in fqns]
```
AliTajeldin: This will probably fail now that the `fqn` PR was merged.

PR #1533, opened 2018-12-03
 #1521 removed needToRun

### LINKED ISSUE #1521: Drop needsToRun from get_graph_json method in smvapp

ninjapapa:
Fixed #1521 

Finally get it removed.

### DIFF
--- a/src/main/python/smv/smvappinfo.py
+++ b/src/main/python/smv/smvappinfo.py
@@ -65,6 +65,7 @@ class SmvAppInfo(object):
 
     def create_graph_json(self):
         """Create dependency graph Json string
+            Dependency graph does not have modules' state info
         """
         (nodes, edges) = self._graph()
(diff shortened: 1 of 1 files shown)


src/main/python/smv/smvappinfo.py
```diff
@@ -80,7 +80,6 @@ def node_dict(n):
                 "fqn": n.fqn(),
                 "type": node_type(n),
                 "version": n.version(),
-                "needsToRun": n.needsToRun(),
```
cloudysail: we can remove the TODO above

PR #1534, opened 2018-12-05
#1498: added get_providers_by_prefix method

### LINKED ISSUE #1498: provide something similar to Java Service Provider API in SMV

AliTajeldin:
fixes #1498 

added the public `get_providers_by_prefix` method which should be used by clients to get list of available providers.

### DIFF
--- a/src/main/python/smv/provider.py
+++ b/src/main/python/smv/provider.py
@@ -29,3 +29,22 @@ class SmvProvider(object):
     @staticmethod
     def provider_type(): return ""
 
+    @classmethod
+    def provider_type_fqn(cls):
+        """create a hierarchichal provider type fqn for a given provider class based on the 
+           provider class hierarchy.
+
+           Example (assume `provider_type()` for class X is X):
+             class A(SmvProvider)
+             class B(A)
+             class C(B)
+           In the above example, C's provider type is just `C` but C's provider_type_fqn is "A.B.C" 
+        """
+        # TODO: handle case where we have multiple inheretence with diamond
+        # (IS_PROVIDER would be true for non-provider in the hierarchy in above case)
+        fqn_parts = [c.provider_type() for c in cls.__mro__ if hasattr(c, "IS_PROVIDER")]
+
+        # actual fqn is in reverse of mro traversal and we can ignore "" type at base provider
+        fqn_parts.reverse()
+        fqn_parts = [f for f in fqn_parts if f != ""]
+        return ".".join(fqn_parts)
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,2 @@
+[MASTER]
+init-hook=sys.path.append('src/main/python')
--- a/src/main/python/smv/datasetrepo.py
+++ b/src/main/python/smv/datasetrepo.py
@@ -233,8 +233,14 @@ class DataSetRepo(object):
 
         return fqns
 
-    def all_providers(self):
+    def _all_providers(self):
+        """scans user libraries and smv libraries for "provider" classes.
+            Returns list of discovered provider classes
+        """
         def is_provider(klass):
+            """A class is a provider if it has `IS_PROVIDER` and is not the base `SmvProvider`
+               which returns empty string for provider type.
+            """
             try:
                 klass_is_provider = (klass.IS_PROVIDER is True) and (klass.provider_type())
             except AttributeError:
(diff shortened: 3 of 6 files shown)


ninjapapa:
Sounds like Python3 does not support `has_key`. 

A question, to apply on connection, is the following the right way?
* Implement `ConnectionProvider(SmvProvider)` - can it be in a sub-package of smv or have to directly under smv folder?
* Implement `SmvJdbcConnectionInfo` - its `provider_type_fqn` will be `conn.jdbc`
* In `get_connection_by_name` method of iomod modules should call `get_providers_by_prefix('conn')` and lookup `jdbc` to get the class
* Use conf props to init `SmvJdbcConnectionInfo`

AliTajeldin:
Reply to above:
* used `in` instead of `has_key` to work with later version of python
* Providers can be in sub-package of user lib or smv lib.  We use tree walk to discover modules.
* correct, `SmvJdbcConnectionInfo` fqn would be `conn.jdbc`
* To lookup `conn.jdbc` you can just search for `conn.jdbc` (should only be a single match).  If you want to find all connection types you would search for `conn.` (note `.` at end).  You would get the class instance which you can use to instantiate the object.  We may want to add a `find_provider_by_name` that would do an exact match on a provider rather than prefix so you can just call `find_provider_by_name('conn.jdbc')`
* correct on using conf for `SmvJdbcConnectionInfo`.  The provider just finds the class, how you init the class and where it gets it conf is up to the class implementer.  For connection info classes we would just use the conf (we would also use the conf to get the connection type as well, `jdbc` in this example).

src/main/python/smv/provider.py
```diff
@@ -29,3 +29,22 @@ class SmvProvider(object):
...
+
+           Example (assume `provider_type()` for class X is X):
+             class A(SmvProvider)
+             class B(A)
+             class C(B)
+           In the above example, C's provider type is just `C` but C's provider_type_fqn is "A.B.C" 
+        """
+        # TODO: handle case where we have multiple inheretence with diamond
+        # (IS_PROVIDER would be true for non-provider in the hierarchy in above case)
+        fqn_parts = [c.provider_type() for c in cls.__mro__ if hasattr(c, "IS_PROVIDER")]
```
laneb: This could get a little weird in case of user error. Suppose above that the user forgot to override `provider_type` for B. They'd end up getting an FQN of `A.A.C`.

AliTajeldin:
@ninjapapa tests passed, ready for re-review

src/main/python/smv/provider.py
AliTajeldin: yah, we are just punting on this until it becomes a problem in real life.  The use cases we have in mind now all do not have a diamond inheritance.

laneb: I'm not following how this relates to diamond inheritance. The inheritance pattern above is linear (`A`<-`B`<-`C`).

AliTajeldin: Aaah, I assumed for linear hierarchy, we can ensure that all classes along the way are indeed providers.  If the user doesn't do that, it is a bug in their code.  Actually, this would be true for diamond as well.  If you mix-in two providers, they should both provide `provider_type()` anyhow.
I'll remove the TODO in another ticket.

PR #1548, opened 2018-12-11
#1546: remove install of r-base

### LINKED ISSUE #1546: smv release script fails on `apt-key`
We had issues with `pgp.mit.edu` in the past but they were resolved within hours.  This has been failing for two days.  The `apt-key` command finally worked today in an ubuntu docker image, but the release script still failed.

```
Step 27/36 : RUN echo "deb http://cran.rstudio.com/bin/linux/debian lenny-cran/" >> /etc/apt/source.list &&    apt-key adv --keyserver hkp://pgp.mit.edu --recv-key 381BA480 &&    apt-get update &&    apt-get install -y r-base &&    apt-get install -y sudo &&    apt-get install -y vim &&    apt-get install -y git &&    apt-get install -y graphviz
 ---> Running in e9f504214a58
Warning: apt-key output should not be parsed (stdout is not a terminal)
Executing: /tmp/apt-key-gpghome.LGkkOmoZHO/gpg.1.sh --keyserver hkp://pgp.mit.edu --recv-key 381BA480
gpg: keyserver receive failed: No data
The command '/bin/sh -c echo "deb http://cran.rstudio.com/bin/linux/debian lenny-cran/" >> /etc/apt/source.list &&    apt-key adv --keyserver hkp://pgp.mit.edu --recv-key 381BA480 &&    apt-get update &&    apt-get install -y r-base &&    apt-get install -y sudo &&    apt-get install -y vim &&    apt-get install -y git &&    apt-get install -y graphviz' returned a non-zero code: 2
make: *** [docker_smv] Error 2
```

AliTajeldin:
Fixes #1546 

* Removed install of `r-base` and associated troublesome `apt-key`.
* Also moved all the `apt-get install` to before the copy of SMV source.  The build image is only used for building SMV so no need.  And moving it before the copy will make that docker layer immune to SMV code changes.

### DIFF
--- a/docker/smv/Dockerfile
+++ b/docker/smv/Dockerfile
@@ -31,7 +31,7 @@ RUN tar -xzvf sbt-$SBT_VERSION.tgz
 RUN mv sbt /usr/lib/sbt
 RUN rm sbt-$SBT_VERSION.tgz
 
-RUN apt-get -y install make gcc libssl-dev zlib1g-dev
+RUN apt-get -y install make gcc libssl-dev zlib1g-dev sudo git vim graphviz
 
 COPY . ${SMV_HOME}
 WORKDIR ${SMV_HOME}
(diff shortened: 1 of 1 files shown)


laneb:
As discussed need to leave the second apt-get line in place so that those tools are available in the final docker image. Otherwise lgtm.

AliTajeldin:
Approved by @laneb (looking over my shoulder) 😄

PR #1552, opened 2018-12-13
I1551 output always run

### LINKED ISSUE #1551: Should output module always run?
Currently output module re-run the same way as other modules, so basically if there is a persisted data with the same hash-of-hash, it will NOT rerun. However, it may not be the desired behavior since user may somehow deleted the output file/table, and expect rerun the output module will recover it.

ninjapapa:
Fixed #1551


src/main/python/smv/iomod/base.py
```diff
@@ -107,6 +107,16 @@ class SmvOutput(SmvIoModule):
     def dsType(self):
         return "Output"
 
+    def isEphemeral(self):
+        """SmvOutput are always NON-ephemeral"""
+
+        # By making SmvOutput NON-ephemeral, it will not cached in the app level
```
cloudysail: I think SMV should not internally persist any data for the output type, which means `isEphemeral` should be true for output.

PR #1553, opened 2018-12-14
I1551 output always run - simpler version

### LINKED ISSUE #1551: Should output module always run?

ninjapapa:
Fixed #1551 

A simpler approach than PR #1552

### DIFF
--- a/src/main/python/smv/smvgenericmodule.py
+++ b/src/main/python/smv/smvgenericmodule.py
@@ -278,7 +278,7 @@ class SmvGenericModule(ABC):
         """
         if (forceRun or (self.versioned_fqn not in self.smvApp.data_cache)):
             res = self._computeData(fqn2df, run_set, collector, is_quick_run)
-            if (self.isEphemeral() and not self.isSmvOutput()):
+            if (self.isEphemeral()):
                 # Only cache ephemeral modules data, since non-ephemeral any how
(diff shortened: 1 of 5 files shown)


src/main/python/smv/smvgenericmodule.py
```diff
@@ -278,14 +278,16 @@ def _get_data(self, fqn2df, run_set, collector, forceRun, is_quick_run):
         """
         if (forceRun or (self.versioned_fqn not in self.smvApp.data_cache)):
             res = self._computeData(fqn2df, run_set, collector, is_quick_run)
-            if (self.isEphemeral()):
+            if (self.isEphemeral() and not self.isSmvOutput()):
```
AliTajeldin: out of curiosity, why not just make output ephemeral like input?

ninjapapa: Output is ephemeral. 

The logic here is `isEphemeral` but not output. Since in general ephemeral will be cached, but we don't want to cache output.

ninjapapa:
Close to move real output operation into a "post_run" method.

ninjapapa:
@AliTajeldin ready to review.

PR #1557, opened 2018-12-18
 #1556 fixed test bug on python 3

### LINKED ISSUE #1556: itest failed on python 3 in PR #1555
PR #1555 used a relative path for import which Python 3 doesn't support. Need to fix.

ninjapapa:
Fixed #1556 

Please wait for CI.

### DIFF
--- a/src/test/python/testDataSetRepo/provider/library/dsalib/__init__.py
+++ b/src/test/python/testDataSetRepo/provider/library/dsalib/__init__.py
@@ -1,1 +1,1 @@
-from d import *
+from dsalib.d import *


src/test/python/testDataSetRepo/provider/library/dsalib/__init__.py
```diff
@@ -1 +1 @@
-from d import *
+from dsalib.d import *
```
AliTajeldin: Does this work in 2.7?

ninjapapa: Yes.

PR #1570, opened 2019-01-10
#1569 set smv lock to true for blank project init

### LINKED ISSUE #1569: set smv.lock to true for blank project init

cloudysail:
Fixes #1569


cloudysail:
@AliTajeldin, in SMV v2r5 and later, the default behavior is `lock = false` when persisting data. To get back the locking mechanism, need to set `smv.lock` to `true`.

PR #1577, opened 2019-01-18
#1572: initial ci-latest-mgr script

### LINKED ISSUE #1572: create `admin/ci-latest.sh` script to upload/download latest ci build result
Need a place to store/download the latest CI build result of SMV.  We will utilize a single "release" to store the latest successfully CI built SMV image.  The script should not depend on any other scripts in SMV repo so that we can just do a `curl ... | bash` to get latest SMV binary build.  We will need to supply the GITHUB user credential on upload as an environment variable.

AliTajeldin:
Fixes #1572

### DIFF
--- a/admin/ci-latest-mgr
+++ b/admin/ci-latest-mgr
@@ -1,0 +1,63 @@
+#!/bin/bash
+#
+# script to upload/download latest SMV release bundle created by CI
+#
+# USAGE:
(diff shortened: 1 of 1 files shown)


AliTajeldin:
yes.  SemaphoreCI does not allow cache across projects.  So we need to build SMV and upload it when SMV is built/tested and then download it when it is needed in another project.

PR #1581, opened 2019-01-19
I1576 content list

### LINKED ISSUE #1576: Need a get_content_list method for connections
Client code:
```python
my_conn = smvApp.get_connection_by_name('my_hdfs')
csv_files = my_conn.get_content_list(pattern=r".*\.csv", ignore_case=True)
```

ninjapapa:
Fixed #1576


src/main/python/smv/conn/conninfos.py
```diff
@@ -32,6 +32,30 @@ def attributes():
...
+        if (self.password is not None):
+            builder = builder.option('password', self.password)
+
+        return builder
+
+    def get_contents(self, smvApp):
+        """Return a list of table names
+        """
+        tables_df = self._connect_for_read(smvApp)\
+            .option("dbtable", "(SELECT TABLENAME FROM SYS.SYSTABLES WHERE TABLETYPE='T') foo")\
```
AliTajeldin: Not sure this is standard.  Will work with microsoft sql server, but I think it will fail on other databases.
For jdbc, you usually have to get the table list from the the jdbc connection meta data (which is standard).
Couldn't find something similar using python DB API

ninjapapa: According to this:
https://stackoverflow.com/questions/42880119/how-to-list-all-tables-in-database-using-spark-sql

```
If your remote DB has a way to query its metadata with SQL, such as INFORMATION_SCHEMA.TABLE (Postgres) or INFORMATION_SCHEMA.TABLES (MySQL, SQL Server) or SYS.ALL_TABLES (Oracle), then you can just use it from Spark to retrieve the list of local objects that you can access.
```

Any suggestion?

AliTajeldin: Basically, you can query the "special" tables but you need to do a different sql command for each database.  The JDBC protocol takes care of that for you using metadata interface.  However, even with that, not all DBs conform to it, but at least the common ones do.

I can think of two options:
1. implement a different connection interface for each database type and would support a couple of common DB (e.g. oracle, mysql, postgress, mssql).  We would then use a different catalog table as shown in the above example.
2. Use a native Java JDBC connection to get the metadata.

My preference is for option 1 and we can start with just one db for now.

ninjapapa: For now, we will return [] for JDBC connections. Will handle the situation in a different issue.

PR #1582, opened 2019-01-19
I1579 conn conf

### LINKED ISSUE #1579: Support config files in a folder
Currently we supports 4 type of configurations other than the default values:

Priority: Low to High
 - default
 - conf/smv-app-conf.props
 - ${HOME}/.smv/smv-user-conf.props
 - conf/smv-user-conf.props
 - command-line

As adding the connection concept, to avoid making existing config files too complicated, we will support config files in folders. A structure this issue should support is:

- conf
   - smv-app-conf.props
   - smv-user-conf.props
   - connections
         - all_hdfs_conns.props
         - all_hive_conns.props

And all config files in folders are at the same priority level as `smv-app-conf.pros`.

ninjapapa:
Fixed #1579

### DIFF
--- a/src/main/python/smv/smvconfig.py
+++ b/src/main/python/smv/smvconfig.py
@@ -33,16 +33,20 @@ class SmvConfig(object):
         self._jvm = _jvm
         self.cmdline = self._create_cmdline_conf(arglist)
 
+        DEFAULT_SMV_APP_CONF_FILE  = "conf/smv-app-conf.props"
+        DEFAULT_SMV_CONN_CONF_FILE  = "conf/connections.props"
+        DEFAULT_SMV_USER_CONF_FILE = "conf/smv-user-conf.props"
         DEFAULT_SMV_HOME_CONF_FILE = os.getenv('HOME', '') + "/.smv/smv-user-conf.props"
 
         self.app_dir = self.cmdline.pop('smvAppDir')
         self.static_props = {}
         self.dynamic_props = {}
 
-        self.app_conf_path = self.cmdline.pop('smvAppConfFile')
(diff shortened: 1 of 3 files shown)


src/main/python/smv/smvconfig.py
```diff
@@ -194,12 +195,14 @@ def _create_cmdline_conf(self, arglist):
...
         DEFAULT_SMV_APP_CONF_FILE  = "conf/smv-app-conf.props"
+        DEFAULT_SMV_CONN_CONF_FILE  = "conf/connections.props"
         DEFAULT_SMV_USER_CONF_FILE = "conf/smv-user-conf.props"
 
         # Where to find props files
         parser.add_argument('--smv-app-dir', dest='smvAppDir', default=".", help="SMV app directory")
-        parser.add_argument('--smv-app-conf', dest='smvAppConfFile', default=DEFAULT_SMV_APP_CONF_FILE, help="app level (static) SMV configuration file path")
-        parser.add_argument('--smv-user-conf', dest='smvUserConfFile', default=DEFAULT_SMV_USER_CONF_FILE, help="user level (dynamic) SMV configuration file path")
+        parser.add_argument('--smv-app-conf', dest='smvAppConfFile', default=DEFAULT_SMV_APP_CONF_FILE, help="app level SMV configuration file path")
+        parser.add_argument('--smv-conn-conf', dest='smvConnConfFile', default=DEFAULT_SMV_CONN_CONF_FILE, help="app level SMV connection configuration file path")
```
AliTajeldin: Should we even make these configurable?  We actually never change the name of the user/app conf files.  If we keep them as configurable, we should also add them to the doc.

PR #1612, opened 2019-04-10
#1611 support file based runtimeConfig and kernelConf

### LINKED ISSUE #1611: support file based runtime config and kernel config
- support `smv.runtimeConfigFile` which defines the relative path against AppDir and defines runtime config props
- support `smv.kernelConfigFile` which defines spark conf props to create Spark session in `SmvDriver`

cloudysail:
Fixes #1611

### DIFF
--- a/src/main/python/smv/smvapp.py
+++ b/src/main/python/smv/smvapp.py
@@ -121,7 +121,7 @@ class SmvApp(object):
 
         self.smvSchemaObj = self._jvm.SmvPythonHelper.getSmvSchema()
 
-        self.py_smvconf = SmvConfig(arglist, self._jvm)
+        self.py_smvconf = SmvConfig(arglist)
 
         # configure spark sql params
         if (self.sparkSession is not None):
--- a/src/main/python/smv/smvconfig.py
+++ b/src/main/python/smv/smvconfig.py
@@ -265,6 +265,12 @@ class SmvConfig(object):
         full_user_conf_path = os.path.join(self.app_dir, self.user_conf_path)
 
         app_conf_props = SmvConfig.load(full_app_conf_path)
+        # runtime_config_file is the relative path which defined in app_conf_props
+        runtime_config_file = app_conf_props.get('smv.runtimeConfigFile')
+        runtime_conf_props_from_file = {}
+        if (runtime_config_file is not None):
+            full_runtime_config_file_path = os.path.join(self.app_dir, runtime_config_file)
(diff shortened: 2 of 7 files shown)


src/main/python/smv/smvapp.py
```diff
@@ -121,7 +121,7 @@ def __init__(self, arglist, _sparkSession, py_module_hotload=True):
 
         self.smvSchemaObj = self._jvm.SmvPythonHelper.getSmvSchema()
 
-        self.py_smvconf = SmvConfig(arglist, self._jvm)
+        self.py_smvconf = SmvConfig(arglist)
```
ninjapapa: So nice we don't have dependency to _jvm now.

src/main/python/smv/smvconfig.py
```diff
@@ -236,6 +246,13 @@ def parse_props(prop):
         res = vars(parser.parse_args(arglist))
         return res
 
+    @staticmethod
```
ninjapapa: Nice!

src/main/python/smv/smvconfig.py
```diff
@@ -236,6 +246,13 @@ def parse_props(prop):
...
+    def load(path):
```
ninjapapa: Nice!

src/main/python/smv/smvconfig.py
```diff
@@ -270,13 +280,20 @@ def load(path):
...
         #   - conf/smv-app-conf.props
+        #   - props defined in the runtime config file
         #   - conf/connections.props
         #   - ${HOME}/.smv/smv-user-conf.props
         #   - conf/smv-user-conf.props
         #   - command-line
         res = {}
         res.update(default_props)
         res.update(app_conf_props)
+        # inject the props defined in the runtime config file if exists
```
AliTajeldin: lets keep the same old pattern of reading props above and merging them with res here.  Push the check for file exists to the `load` method instead.  So code here should just have
```
runtime_conf_props_from_file = SmvConfig.load(...) # above with rest of loads
...
res.update(runtime_conf_props_from_file) . # right here
```

cloudysail: hmmm...I did not do that because this `runtime_conf_props_from_file` depends on the updated `res` after reading the `app_conf_props`.

cloudysail: Aha, figured out a cleaner way to do this and ready for review now.

